PR #1, opened 2015-12-17
Blocking futures support

agrippa:
This diff adds support for future-esque asyncs and blocking on DDFs, in both the C and C++ APIs. In particular, it adds the following functions to the C API:

```
/*
 * Block the currently executing task on the provided DDF. Returns the datum
 * that was put on ddf.
 */
void *hclib_ddf_wait(hclib_ddf_t *ddf);
/*
 * Spawn an async that automatically puts a DDF on termination.
 */
hclib_ddf_t *hclib_async_future(generic_framePtr fp, void *arg,
        hclib_ddf_t **ddf_list, struct _phased_t *phased_clause,
        int property);
```

hclib_async_future is the simpler to implement, we just allocate a DDF and wrap the user function so that we can do a ddf_put on completion.

hclib_ddf_wait uses the lightweight context work to swap off the current context on a wait, entering the main work-stealing loop instead. This implementation is based on the support needed to block on end-finishes.

I've also added equivalent C++ APIs, and added a couple of tests for both APIs (future0 and future1).

This diff is based off of merge, so I'll merge this stuff into merge before merging merge into master (now that sentence wasn't hard to follow was it?).

### DIFF
--- a/inc/hclib.h
+++ b/inc/hclib.h
@@ -84,8 +84,7 @@ void hclib_async(asyncFct_t fct_ptr, void * arg,
         int property);
 
 /*
- * Spawn an async that automatically puts a DDF on termination. The put is
- * performed with arg.
+ * Spawn an async that automatically puts a DDF on termination.
  */
 hclib_ddf_t *hclib_async_future(generic_framePtr fp, void *arg,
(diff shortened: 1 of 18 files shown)


agrippa:
Merged into master locally.

PR #2, opened 2015-12-17
Nonblocking finish

agrippa:
This review adds the concept of a non-blocking finish and uses it to implement a future-like forasync.

A non-blocking finish can be used to group tasks within a finish scope together. Rather than blocking on an end-finish, a non-blocking finish returns a promise that can be used to synchronize on the completion of all tasks in the finish scope. This allows the chaining of finish scopes and individual asyncs (or other asynchronous constructs in HCLib, such as asynchronous copies to a GPU).

This review also implements an example use of a nonblocking finish, a future-like forasync that returns a single event which can be used to synchronize on all tasks that are part of the same forasync.

Examples/tests are also added for both nonblocking finishes and future-forasync.


agrippa:
Merged into futures locally.

PR #3, opened 2015-12-19
GPU Support v1

agrippa:
1. Unification of the internal spawn API to simplify runtime code: There used to be several variants of spawn, each of them supporting a slightly different feature set (await vs. at vs. future vs. ...). I've put all of that functionality in a single shared function, with the existing spawn APIs sitting on top of it.
2. Allow forasyncs to be created at a place, and for their execution to be dependent on a list of DDFs.
3. Added an HC_CUDA preprocessor definition to indicate whether the project is being built with or without GPU support. CUDA support is added with the --enable-cuda configuration flag.
4. A dedicated GPU thread, very similar to the communication worker.
5. Memory allocation, release, clear, and copying routines for a HPT with separate address spaces. See allocate_at, free_at, memset_at, and async_copy.
6. New tests under the test/cuda directory as well as an nbody example.
7. Miscellaneous style cleanup.


inc/hclib_cpp.h
```diff
@@ -31,6 +31,38 @@ int current_worker();
...
 void get_places(place_t **pls, place_type_t type);
+place_t *get_current_place();
+place_t **get_children_of_place(place_t *pl, int *num_children);
+place_t *get_root_place();
+place_t **get_nvgpu_places(int *n_nvgpu_places);
+char *get_place_name(place_t *pl);
+
+#ifdef HC_CUDA
+template<typename T>
+T *allocate_at(place_t *pl, size_t nitems, int flags) {
```
vkaries, line 45: In Habanero-C/hclib, the data driven task/future (blocking constructs) syntax was "DDF_FREE", "DDF_CREATE", DDF_HANDLE", "HC_MALLOC", "HC_FREE", etc. If we follow the same naming convention then we should change all hcpp's synchronous (blocking constructs) as "ALLOC_AT", "DEALLOC_AT" (or FREE_AT), MEMSET_AT.. etc.... ?

inc/hcpp-async.h
```diff
@@ -163,6 +164,13 @@ inline void _asyncAwait(hclib_ddf_t ** ddf_list, T lambda) {
 }
 
 template <typename T>
+inline void asyncAwaitAt(hclib_ddf_t **ddf_list, place_t *pl, T lambda) {
```
vkaries, line 167: How about having place_t as first parameter and ddf arguments following this. In case of C++ interface, the user will pass "N" number of DDFs as arguments and it will more readable if they make the first parameter as place_t followed by comma separated DDFs

inc/hclib_cpp.h
vkaries, line 45: Or maybe "HC_MALLOC_AT", HC_FREE_AT"

src/Makefile.am
```diff
@@ -9,15 +9,28 @@ CXXFLAGS = -Wall -g -O3 -std=c++11
 LDFLAGS = -lpthread
 
 if DIST_WS
-HUPCPP_FLAGS_1 = -DDIST_WS
+HC_FLAGS_1 = -DDIST_WS
```
vkaries, line 12: I think this would confuse reading the code. HUPCPP_\* naming convention make it easier to read. Next when we will have HCOpenSHMM, then there will be another flag as HOSHMM_*, making easier to understand without coming back to this file to see what it means.

src/hclib.c
```diff
@@ -68,9 +79,10 @@ hclib_ddf_t *hclib_async_future(futureFct_t fp, void *arg,
 #define DEBUG_FORASYNC 0
 
 forasync1D_task_t * allocate_forasync1D_task() {
-    forasync1D_task_t * forasync_task = (forasync1D_task_t *) malloc(
+    forasync1D_task_t *forasync_task = (forasync1D_task_t *) malloc(
             sizeof(forasync1D_task_t));
     assert(forasync_task && "malloc failed");
```
vkaries: Sorry if I left some places still using assert rather than HC_ASSERT. This should be changed at all places.

src/hclib.c
```diff
@@ -329,9 +343,9 @@ void forasync1D_flat(void * forasync_arg) {
         printf("Scheduling Task %d %d\n",low0,high0);
```
vkaries, line 342: Probably you missed to remove this print statement.

src/hclib.c
```diff
@@ -355,9 +369,9 @@ void forasync2D_flat(void * forasync_arg) {
             printf("Scheduling Task %d %d\n",low1,high1);
```
vkaries, line 368: print statement..

src/hclib_cpp.cpp
```diff
@@ -36,6 +36,26 @@ void hclib::get_places(hclib::place_t **pls, hclib::place_type_t type) {
...
+
+place_t **hclib::get_nvgpu_places(int *n_nvgpu_places) {
+    return hclib_get_nvgpu_places(n_nvgpu_places);
+}
+
+char *hclib::get_place_name(place_t *pl) {
+    return hclib_get_place_name(pl);
+}
+
 void *hclib::ddf_wait(hclib::ddf_t *ddf) {
```
vkaries, line 67: I think we should use the same naming convention as Habanero-C and have DDF_WAIT, etc.

src/hcpp-deque.c
```diff
@@ -39,108 +39,108 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-	int n = (deq->tail) % INIT_DEQUE_CAPACITY;
-	deq->data[n] = (task_t*) entry;
-	deq->tail++;
-	return 1;
+int deque_push(deque_t* deq, void* entry) {
+    int size = deq->tail - deq->head;
+    if (size == INIT_DEQUE_CAPACITY) { /* deque looks full */
+        /* may not grow the deque if some interleaving steal occur */
+        // std::cout<<getenv("PMI_RANK") <<": Deque full for worker-"<<current_ws()->id << std::endl;
+        // HASSERT("DEQUE full, increase deque's size " && 0);
```
vkaries, line 55: Can we uncomment this assert? I think in production build, HASSERT doesn't do anything

src/hcpp-hpt.c
```diff
@@ -275,12 +304,163 @@ inline void init_hc_deque_t(hc_deque_t * hcdeq, place_t * pl){
 #endif
 }
 
-/*Initialize deq's buffer (data + capacity)*/
-void init_deq_buffer(hc_workerState * ws, deque_t * deq, int capacity) {
-    //TODO: in our current design this does not makes sense
+#ifdef HC_CUDA
+void *hclib_allocate_at(place_t *pl, size_t nbytes, int flags) {
```
vkaries, line 307: Maybe:
#ifndef CPP
#define HC_MALLOC_AT hclib_allocate_at
#else
#define HC_MALLOC_AT allocate_at
#endif

src/hcpp-hpt.c
```diff
@@ -275,12 +304,163 @@ inline void init_hc_deque_t(hc_deque_t * hcdeq, place_t * pl){
...
+    assert(pl); assert(nbytes > 0);
```
vkaries: HASSERT

src/hcpp-hpt.c
```diff
@@ -275,12 +304,163 @@ inline void init_hc_deque_t(hc_deque_t * hcdeq, place_t * pl){
...
+            } else {
+                hclib_memory_tree_insert(ptr, nbytes,
+                        &hcpp_context->pinned_host_allocs);
+                return ptr;
+            }
+        }
+#else
+        HASSERT(flags == NONE);
+#endif
+        return malloc(nbytes);
```
vkaries, line 339: HC_MALLOC

src/hcpp-hpt.c
```diff
@@ -275,12 +304,163 @@ inline void init_hc_deque_t(hc_deque_t * hcdeq, place_t * pl){
...
+    return hclib_memory_tree_contains(ptr, &hcpp_context->pinned_host_allocs);
+}
+
+void hclib_free_at(place_t *pl, void *ptr) {
+    if (is_cpu_place(pl)) {
+        if (is_pinned_cpu_mem(ptr)) {
+            hclib_memory_tree_remove(ptr, &hcpp_context->pinned_host_allocs);
+            CHECK_CUDA(cudaFreeHost(ptr));
+        } else {
+            free(ptr);
```
vkaries, line 373: HC_FREE

src/hcpp-hpt.c
```diff
@@ -275,12 +304,163 @@ inline void init_hc_deque_t(hc_deque_t * hcdeq, place_t * pl){
...
+    } else {
+        spawn_gpu_task((task_t *)task);
+    }
+
+    return ddf;
+}
+
+hclib_ddf_t *hclib_async_memset(place_t *pl, void *ptr, int val, size_t nbytes,
+        hclib_ddf_t **ddf_list, void *user_arg) {
+    gpu_task_t *task = malloc(sizeof(gpu_task_t));
```
vkaries, line 431: HC_MALLOC

src/hcpp-hpt.c
```diff
@@ -535,22 +738,28 @@ place_t * parsePlaceElement(xmlNode * plNode) {
...
             tmp->pl = pl;
             if (pl->workers == NULL) pl->workers = tmp;
             else wslast->next_worker = tmp;
             wslast = tmp;
+            nchildren++;
         }
         child = child->next;
     }
+
+    if (type == NVGPU_PLACE) assert(nchildren == 0);
```
vkaries: HASSERT

src/hcpp-runtime.c
```diff
@@ -45,16 +45,24 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "hcpp-finish.h"
 #include "hcpp-hpt.h"
 #include "hcupc-support.h"
+#include "hcpp-cuda.h"
 
 // #define VERBOSE
 
+static const int communication_worker_id = 1;
```
vkaries: Does this mean, the communication worker id = 1 and not 0 ? If yes, then this would affect HabaneroUPC++ significantly

src/hcpp-runtime.c
```diff
@@ -166,10 +180,15 @@ void hcpp_global_init() {
         hcpp_context->done_flags[i].flag = 1;
     }
 
-#ifdef HCPP_COMM_WORKER
-    comm_worker_out_deque = (semiConcDeque_t *)malloc(sizeof(semiConcDeque_t));
```
vkaries, line 170: HC_MALLOC

src/hcpp-runtime.c
```diff
@@ -246,11 +265,21 @@ void hcpp_entrypoint() {
...
-#ifdef HCPP_COMM_WORKER
-    starting_worker = 2;
+
+    for (int i = 1; i < hcpp_context->nworkers; i++) {
+#ifdef HC_COMM_WORKER
+        /*
+         * If running with a thread dedicated to network communication (e.g. through
+         * UPC, MPI, OpenSHMEM) then skip creating that thread as a compute worker.
+         */
+        assert(communication_worker_id > 0);
```
vkaries: Again, same question, whether you are changing communication worker id? If yes then this needs to be discussed why its important

src/hcpp-runtime.c
```diff
@@ -246,11 +265,21 @@ void hcpp_entrypoint() {
...
+        if (i == communication_worker_id) continue;
+#endif
+#ifdef HC_CUDA
+        assert(gpu_worker_id > 0);
```
vkaries: HASSERT

test/cuda/Makefile
```diff
@@ -0,0 +1,16 @@
+include $(HCPP_ROOT)/include/hcpp.mak
+
+TARGETS=allocate copy kernel
+
+FLAGS=-g -std=c++11
```
vkaries, line 5: Can we make this FLAG as part of PROJECT_CFLAGS?

test/cuda/Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+all: $(TARGETS)
+
+kernel: kernel.cu
+	nvcc -arch=sm_30 ${FLAGS} $(PROJECT_CFLAGS) $(PROJECT_LDFLAGS) -o $@ $^ $(PROJECT_LDLIBS)
```
vkaries, line 10: nvcc can be defined in hcpp.mak as CXX_GPU = nvcc -arch=sm_30

inc/hclib_cpp.h
agrippa, line 45: I think this came up in the last review too. If we're standardizing on hclib, we want to give consistently named functions that make it clear which package they belong to. For example, when you write an MPI program you don't call Send, Recv, Barrier, etc. You call MPI_Send, MPI_Receive, etc. The function names are always prefixed with the package name to prevent conflicts with other libraries. We're doing the same thing with the C API, so we'll always have hclib_allocate_at instead of just allocate_at. The C++ API is already exactly as you said, but in the namespace hclib.

If you were talking about the capitalization, those things were capitalized in the old HC because they were usually preprocessor macros, and that's the convention for macros. It wasn't a great idea, but I'm sure there was a good reason at the time. We shouldn't do that here, these are functions, typing in all caps is a pain, and I don't think there's a single standard API out there that uses all capitalized letters in their naming conventions.

src/Makefile.am
agrippa, line 12: I would disagree. Right below:

if HUPCPP
HC_FLAGS_2 = -DHUPCPP -DHC_COMM_WORKER
else
HC_FLAGS_2 =
endif

It is very clear from the conditional of that if statement that these flags relate to HUPCPP. Even more importantly, they relate to HUPCPP but are applied to HC. The conditional tells you what the flag relates to or is controlled by, the variable name tells you what they apply to or are used to build. Just like below we have:

if X86
if OSX
libhcpp_la_SOURCES += \
fcontext/jump_i386_sysv_macho_gas.S \
fcontext/make_i386_sysv_macho_gas.S
else

It's clear that the thing controlling the variable being set is the platform we're running on (x86, OSX), but that the variable being set applies to some source code for "libhcpp". If instead you did:

x86_la_SOURCES += \
fcontext/jump_i386_sysv_macho_gas.S \
fcontext/make_i386_sysv_macho_gas.S

you actually end up conveying less information to the reader.

src/hclib.c
agrippa: k. i've replaced all calls to assert with HASSERT

src/hclib.c
agrippa, line 342: Huh? I didn't add that print statement. In any case, it's protected by DEBUG_FORASYNC so it is normally disabled.

src/hclib.c
agrippa, line 368: Same comment, see above. Things I add are highlighted in green, things I delete are highlighted in red. This isn't highlighted in either, so it is unchanged in this diff.

src/hcpp-deque.c
agrippa, line 55: I don't know. That assert was already commented, I haven't changed anything to do with that.

src/hclib_cpp.cpp
agrippa, line 67: Do you mean you want to capitalize everything? See my comments above on why DDF_WAIT was originally capitalized, and why that's a bad idea. I don't think I can name a single API that uses capitalization as its style (except for maybe fortran or sequel stuff, certainly not C/C++ stuff).

src/hcpp-hpt.c
agrippa: fixed.

src/hcpp-hpt.c
agrippa, line 307: What are you commenting on? How does that macro help?

src/hcpp-hpt.c
agrippa, line 339: allocate_at should not replace internal memory allocations. That is purely a user-facing API. Using it ourselves would just add unnecessary overheads.

src/hcpp-hpt.c
agrippa, line 373: I don't like this idea of wrapping things in macros, and capitalizing everything. Can you explain your rationale for these changes?

src/hcpp-hpt.c
agrippa, line 431: Same comment. Single word comments in code reviews make it hard for people to fix problems in the code, because they have no way to understand the reasoning behind them.

src/hcpp-hpt.c
agrippa: fixed.

src/hcpp-runtime.c
agrippa: In what way?

src/hcpp-runtime.c
agrippa, line 170: Needs explanation.

src/hcpp-runtime.c
agrippa: It simplifies the code. Thread 0 is the main entry thread, spawned by the system. Thread 1 is the communication thread. Thread 2 is the GPU thread. I don't see anywhere in your code where you assume that Thread 0 is the communication thread (and if that is the case, it's probably something that should be factored out anyway). Again, can you explain how this breaks anything?

src/hcpp-runtime.c
agrippa: fixed.

test/cuda/Makefile
agrippa, line 5: This is for a particular example. We wouldn't want it to apply to all hclib projects, everywhere.

test/cuda/Makefile
agrippa, line 10: It could be, but it shouldn't. Forcing -arch=sm_30 means that there are a lot of GPUs this wouldn't even compile for, but which we could support.

PR #9, opened 2016-05-31
Minimize context swaps

DaoWen:
This optimization helps to avoid unnecessary creation of new fibers when the workers aren't doing too much stealing. This change brought the number of fiber-swaps in the Fib(35) test down from over 14 million to just a few hundred.

This should give better performance in many cases, but it's still possible to create a scenario where you get a new fiber per finish scope.

### DIFF
--- a/src/hclib-runtime.c
+++ b/src/hclib-runtime.c
@@ -159,10 +159,10 @@ void hclib_global_init() {
                                   &hclib_context->workers, &hclib_context->nworkers);
 
 #ifdef HC_COMM_WORKER
-    assert(hclib_context->nworkers > COMMUNICATION_WORKER_ID);
+    HASSERT(hclib_context->nworkers > COMMUNICATION_WORKER_ID);
 #endif
 #ifdef HC_CUDA
-    assert(hclib_context->nworkers > GPU_WORKER_ID);
+    HASSERT(hclib_context->nworkers > GPU_WORKER_ID);
 #endif
 
     for (int i = 0; i < hclib_context->nworkers; i++) {
--- a/src/hclib-runtime.c
+++ b/src/hclib-runtime.c
@@ -353,9 +353,9 @@ static inline void check_out_finish(finish_t *finish) {
     if (finish) {
         // hc_atomic_dec returns true when finish->counter goes to zero
         if (hc_atomic_dec(&(finish->counter))) {
-#if HCLIB_LITECTX_STRATEGY
+            #if HCLIB_LITECTX_STRATEGY
             hclib_promise_put(finish->finish_deps[0]->owner, finish);
-#endif /* HCLIB_LITECTX_STRATEGY */
+            #endif /* HCLIB_LITECTX_STRATEGY */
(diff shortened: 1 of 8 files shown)


src/hclib-runtime.c
```diff
@@ -1062,32 +1059,68 @@ void help_finish(finish_t *finish) {
...
-        hclib_promise_t *finish_promise = hclib_promise_create();
-        hclib_future_t *finish_deps[] = { &finish_promise->future, NULL };
-        finish->finish_deps = finish_deps;
         // TODO - should only switch contexts after actually finding work
-        LiteCtx *currentCtx = get_curr_lite_ctx();
-        HASSERT(currentCtx);
-        LiteCtx *newCtx = LiteCtx_create(_help_finish_ctx);
-        newCtx->arg = finish;
+
+        // Try to execute a sub-task of the current finish scope
```
DaoWen, line 1064: This is the key change in this pull request.

src/hclib-runtime.c
```diff
@@ -929,17 +932,13 @@ static void _finish_ctx_resume(void *arg) {
...
-void crt_work_loop(LiteCtx *ctx);
-
 // Based on _help_finish_ctx
 void _help_wait(LiteCtx *ctx) {
     hclib_future_t **continuation_deps = ctx->arg;
     LiteCtx *wait_ctx = ctx->prev;
 
-    hclib_dependent_task_t *task = (hclib_dependent_task_t *)malloc(sizeof(
-                                       hclib_dependent_task_t));
+    hclib_dependent_task_t *task = calloc(1, sizeof(*task));
```
DaoWen, line 940: I think this was copied over from a C++ source file. You don't need to cast the result of `malloc` in C. Also, using `calloc` instead lets us avoid the separate `memset` call.

src/hclib-runtime.c
```diff
@@ -1096,9 +1129,8 @@ void help_finish(finish_t *finish) {
 
 void hclib_start_finish() {
     hclib_worker_state *ws = CURRENT_WS_INTERNAL;
-    finish_t *finish = (finish_t *) HC_MALLOC(sizeof(finish_t));
+    finish_t *finish = malloc(sizeof(*finish));
```
DaoWen, line 1132: We should probably just get rid of the `HC_MALLOC` and `HC_FREE` macros.

agrippa, line 1132: I would be happy with this, I'm not sure what the original motivation of them was.

agrippa:
This diff looks fine to me, thanks for the change Nick! I've got a performance regression suite on my branch, if I have time I'll try to merge this diff in and run the suite to see if any of the benchmarks see an appreciable speedup.

src/hclib-runtime.c
DaoWen, line 1132: The original motivation was to use the Cilk allocator in place of `malloc`—but now we have easy ways to link against TBB malloc and other scalable allocators, so I think it's lost its utility.

PR #13, opened 2016-06-18
Remove old allocator macros

DaoWen:
The HC_MALLOC/FREE macros are an artifact of the legacy HC runtime, which at one time had a modified version of the Cilk allocator in the codebase. We now prefer linking against a 3rd-party scalable allocator (e.g., tbb_malloc), making these macros obsolete.

There are some places where the C++ code is now calling `malloc`/`memcpy`/`free` directly instead of calling `new`/`delete`. I've added FIXME comments to mark these issues, but fixing these will requires some more extensive refactoring.

### DIFF
--- a/scripts/gen-generic-asyncPhased.pl
+++ b/scripts/gen-generic-asyncPhased.pl
@@ -34,8 +34,8 @@ for (my $j=0; $j<$ARGV[0]; $j++) {
 
         my $total = $j + 1;
         print "\tint total = $total;\n";
-        print "\tPHASER_t* phaser_type_arr = (PHASER_t*) HC_MALLOC(sizeof(PHASER_t) * total);\n";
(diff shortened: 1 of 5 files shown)


scripts/gen-generic-asyncPhased.pl
```diff
@@ -34,8 +34,8 @@
 
         my $total = $j + 1;
         print "\tint total = $total;\n";
-        print "\tPHASER_t* phaser_type_arr = (PHASER_t*) HC_MALLOC(sizeof(PHASER_t) * total);\n";
-        print "\tPHASER_m* phaser_mode_arr = (PHASER_m*) HC_MALLOC(sizeof(PHASER_m) * total);\n";
+        print "\tPHASER_t* phaser_type_arr = new PHASER_t[total];\n";
+        print "\tPHASER_m* phaser_mode_arr = new PHASER_m[total];\n";
```
DaoWen, line 39: What is this perl script for? I went ahead and updated it, but do we actually need it?

PR #14, opened 2016-06-20
Single type name for loop domains

DaoWen:
For some reason we were using two different type names for the loop domain struct. See commit message (b1e6fc2) for details.

---

<i>Note: This pull request builds on #13. You can compare the branches corresponding to the pull requests to see the diff between just #13 and this PR: https://github.com/habanero-rice/hclib/compare/cxx-cleanup...cxx-cleanup-2</i>

### DIFF
--- a/inc/hclib-async.h
+++ b/inc/hclib-async.h
@@ -64,7 +64,9 @@ inline void call_lambda(T* lambda) {
 	const int wid = current_ws()->id;
 	MARK_BUSY(wid);
 	(*lambda)();
-	HC_FREE((void*) lambda);
+    // FIXME - this whole call chain is kind of a mess
+    // leaving C malloc/free and memcpy calls for now (come back to fix it later)
+	free(lambda);
 	MARK_OVH(wid);
 }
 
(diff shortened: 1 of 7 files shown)


inc/hclib-async.h
```diff
@@ -64,7 +64,9 @@ inline void call_lambda(T* lambda) {
 	const int wid = current_ws()->id;
 	MARK_BUSY(wid);
 	(*lambda)();
-	HC_FREE((void*) lambda);
+    // FIXME - this whole call chain is kind of a mess
```
agrippa, line 67: There are a couple of these comments, but I'm not clear on what qualifies as a mess? Do you mean it needs more documentation, or is there something specifically objectionable about this region of code? It's fine if there is something ugly here that we leave for later, but we'll need to remember what it was 6 months from now when we get back to it.

agrippa:
Other than my one comment on comments, looks good!

inc/hclib-async.h
DaoWen, line 67: This was just a reminder for me. The issues here are already fixed—but the fixes are further down the pull request pipeline. My vague "mess" description was referring to the following issues:
- Calls to memory-related C functions from C++ code (malloc/free/memset).
- Lambdas passed by-value through this call chain, causing a bunch of superfluous copies. (I have an example in a later revision (yet to be pull-requested) showing how each lambda gets copied 2 or 3 times here, vs no copies after some optimization.)
- Unnecessary levels of indirection around the async and async-await task lambdas.
- Duplication of code that's implemented in the underlying C API, while it should be easy to delegate to the C code (e.g., for task allocation).

There's still one place in the HCUPC++ code that has a comment like this, but I'm going to talk with Kumar this week (after he has a time to review these PRs) to re-sync the UPC-related code with these changes.

agrippa, line 67: Okay, if these issues are already fixed then it's not an issue.

vivkumar, line 67: Nick, maintaining backward compatibility with HCUPC++ might not be easy as hclib codebase has significantly changed (e.g. continuation creation has broken the support of old design of communication worker). Max is working a generic resource worker that he plans as a pluggable module for any communication library (e.g., MPI, UPC++, OpenSHMEM) and I believe he would merge that branch into master soon.

Issue #12, opened 2016-06-14
memcpy non-trivially-copyable function objects


DaoWen:
Using `memcpy` to clone C++ objects that are not [<i>trivially copyable</i>](http://en.cppreference.com/w/cpp/concept/TriviallyCopyable) can cause serious problems.

We're currently using `memcpy` to clone lambdas for `async` (see [inc/hclib-async.h:124](https://github.com/habanero-rice/hclib/blob/25bd6ecd22c528490e02bc1f71fd2383e0d5673e/inc/hclib-async.h#L124-L125)).

Here's a simple program demonstrating the problem:

```
#include <memory>
#include <iostream>
#include <unistd.h>
#include <cassert>

#include "hclib_cpp.h"

struct SimpleObject {
    int value;
    SimpleObject(): value(1) { }
    ~SimpleObject() { value = 0; }
};

int main (int argc, char ** argv) {
    hclib::launch([]() {
        hclib::finish([]() {{
            auto p = std::make_shared<SimpleObject>();
            std::cout << "Value starts as " << p->value << std::endl;
            hclib::async([=](){
                usleep(100);
                std::cout << "Value in async is " << p->value << std::endl;
                assert(p->value == 1);
            });
            // p is dead
        }});
    });
    std::cout << "Exiting..." << std::endl;
    return 0;
}
```

Expected output:

```
Value starts as 1
Value in async is 1
Exiting...
```

Current output:

```
Value starts as 1
Value in async is 189792480
Assertion failed: (p->value == 1), function operator(), file capture0.cpp, line 56.
Abort trap: 6
```

DaoWen:
This issue is fixed for the "core" async code in f8240f0b5070051939e04142bb0df8cbb0f5b25b, but there are probably still issues with the "extension" async support (CUDA and UPC++).

I also added the above program as a new C++ test in dc58a3ab9e1f291e153e8d4c11c5d8c170336718.

This issue should be automatically closed when those commits are merged into master. We can open a separate issue for the CUDA / UPC++ asyncs we haven't resolved the issue before merging the above changes.

PR #15, opened 2016-06-20
Don't memcpy C++ objects

DaoWen:
We should be using C++ copy constructors to copy lambdas into the heap. Bad stuff can happen if we use `memcpy` in C++ (see #12, and the newly-added tests).

I also got rid of two levels of indirection around the lambda used for the async task. I'm pretty sure we didn't need them for anything.

---

<i>Note: This pull request builds on #14. You can compare the branches corresponding to the pull requests to see the diff between just #14 and this PR: https://github.com/habanero-rice/hclib/compare/cxx-cleanup-2...cxx-cleanup-3</i>

### DIFF
--- a/inc/hclib-async.h
+++ b/inc/hclib-async.h
@@ -55,71 +55,25 @@ namespace hclib {
  * TODO optimize that overhead.
  */
 
-/*
- * At the lowest layer in the call stack before entering user code, this method
- * invokes the user-provided lambda.
- */
-template <typename T>
-inline void call_lambda(T* lambda) {
-	const int wid = current_ws()->id;
-	MARK_BUSY(wid);
-	(*lambda)();
-    // FIXME - this whole call chain is kind of a mess
-    // leaving C malloc/free and memcpy calls for now (come back to fix it later)
-	free(lambda);
-	MARK_OVH(wid);
-}
-
-/*
- * Store a reference to the type-specific function for calling the user lambda,
- * as well as a pointer to the lambda's location on the heap (through which we
- * can invoke it). async_arguments is stored as the args field in the task_t
- * object for a task, and passed to lambda_wrapper.
- */
-template <typename Function, typename T1>
-struct async_arguments {
-    Function lambda_caller;
-    T1 lambda_on_heap;
-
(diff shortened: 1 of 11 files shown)


inc/hclib-async.h
```diff
@@ -55,79 +55,30 @@ namespace hclib {
...
-     */
-	T* lambda_on_heap = (T*)HC_MALLOC(lambda_size);
-	memcpy(lambda_on_heap, &lambda, lambda_size);
-
-    hclib_task_t t;
-    initialize_task(&t, call_lambda<T>, lambda_on_heap);
-	memcpy(task, &t, sizeof(hclib_task_t));
-	return task;
+    typedef typename std::remove_reference<T>::type U;
+    // FIXME - this whole call chain is kind of a mess
```
agrippa, line 70: Same comment as before, if there's something to fix we need to be more precise than "a mess".

inc/hclib-async.h
```diff
@@ -55,79 +55,30 @@ namespace hclib {
...
-    t->_fp = lambda_wrapper<Function, T1 *>;
-    t->args = args;
-    t->is_async_any_type = 0;
-    t->future_list = NULL;
-    t->place = NULL;
+template<typename T>
+void lambda_wrapper(void *arg) {
+    T* lambda = static_cast<T*>(arg);
+    MARK_BUSY(current_ws()->id);
+    (*lambda)(); // !!! May cause a worker-swap !!!
```
agrippa, line 62: What does this comment mean?

agrippa:
As long as the tests passed these changes look good to me as well. As far as I can tell, the changes to task indirection should retain the current behavior.

inc/hclib-async.h
DaoWen, line 62: There are similar comments a few places in the C code. This a reminder that you can't reuse worker-specific values after this function call—e.g., `current_ws()->id`—since the task might block inside the user code, resulting in a context switch.

vivkumar, line 62: As Max commented, if tests are running fine, this should be ok. But don't remove the comment you have there currently. This would help in future if this leads to undesirable behavior

PR #16, opened 2016-06-21
Unconstrained promise values

DaoWen:
You should be able to store whatever you want in a promise. The restriction on the promise payload value was carried over from the old HC runtime (although I'm not sure how necessary it was there either). We just check the `wait_list_head` to see if a promise is satisfied. This might add a bit of overhead compared to directly checking the value of `datum`, but I wouldn't expect the difference to be significant (the two fields are probably on the same cache line).

Managing all of the casts back and forth between the "dependent task" type and the normal task type was giving me a headache, so I collapsed both types back into the normal task type. We might want to split them up again later, but having just one task type while doing this refactoring meant one less thing that I had to worry about. Since the overhead of combining the two is just one pointer field, I don't think it will cause too much concern for now.

Both of these changes lay groundwork for changes in the C++ API in later commits, such as template specializations of the promise type, and code consolidation for task management.

There are also some fixes for bugs in the test suite that were uncovered while refactoring.

---

<i>Note: This pull request builds on #15. You can compare the branches corresponding to the pull requests to see the diff between just #15 and this PR: https://github.com/habanero-rice/hclib/compare/cxx-cleanup-3...cxx-cleanup-4</i>

### DIFF
--- a/inc/hclib-promise.h
+++ b/inc/hclib-promise.h
@@ -94,7 +94,7 @@ typedef struct hclib_triggered_task_st {
 // We define a typedef in this unit for convenience
 typedef struct hclib_promise_st {
     hclib_future_t future;
-	int kind;
+    promise_kind_t kind;
     volatile void *datum;
     // List of tasks that are awaiting the satisfaction of this promise
     volatile hclib_triggered_task_t *wait_list_head;
--- a/inc/hclib-promise.h
+++ b/inc/hclib-promise.h
@@ -95,9 +95,9 @@ typedef struct hclib_triggered_task_st {
 typedef struct hclib_promise_st {
     hclib_future_t future;
     promise_kind_t kind;
-    volatile void *datum;
+    void *volatile datum;
     // List of tasks that are awaiting the satisfaction of this promise
-    volatile hclib_triggered_task_t *wait_list_head;
+    hclib_triggered_task_t *volatile wait_list_head;
 } hclib_promise_t;
 
 /**
--- a/inc/hclib-promise.h
+++ b/inc/hclib-promise.h
@@ -73,23 +73,7 @@ typedef struct _hclib_future_t {
     struct hclib_promise_st *owner;
 } hclib_future_t;
 
-/**
- * An hclib_triggered_task_t associates dependent tasks and the promises that
- * trigger their execution. This is exposed so that the runtime knows the size
- * of the struct.
- */
-typedef struct hclib_triggered_task_st {
-    // NULL-terminated list of futures this task is registered on
-    hclib_future_t **waiting_frontier;
-    /*
-     * This allows us to chain all dependent tasks waiting on a same promise.
-     * Whenever a triggered task wants to register on a promise, and that
-     * promise is not ready, we chain the current triggered task and the
-     * promise's wait_list_head and try to cas on the promise's wait_list_head,
-     * with the current triggered task.
-     */
-    struct hclib_triggered_task_st *next_waiting_on_same_future;
-} hclib_triggered_task_t;
+struct hclib_task_t;
 
 // We define a typedef in this unit for convenience
 typedef struct hclib_promise_st {
@@ -97,7 +81,7 @@ typedef struct hclib_promise_st {
     promise_kind_t kind;
     void *volatile datum;
     // List of tasks that are awaiting the satisfaction of this promise
-    hclib_triggered_task_t *volatile wait_list_head;
+    struct hclib_task_t *volatile wait_list_head;
 } hclib_promise_t;
 
 /**
@@ -160,11 +144,5 @@ void hclib_promise_put(hclib_promise_t *promise, void *datum);
  * that was put on promise.
  */
 void *hclib_future_wait(hclib_future_t *future);
-
-/*
- * Some extras
- */
-void hclib_triggered_task_init(hclib_triggered_task_t *task,
-        hclib_future_t **future_list);
 
 #endif /* HCLIB_PROMISE_H_ */
--- a/inc/hclib-task.h
+++ b/inc/hclib-task.h
@@ -22,7 +22,7 @@
  *      this task depends on to execute, and which it will wait on before
  *      running.
  */
-typedef struct _hclib_task_t {
+typedef struct hclib_task_t {
     void *args;
     struct finish_t *current_finish;
     generic_frame_ptr _fp;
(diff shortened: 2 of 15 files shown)


inc/hclib-promise.h
```diff
@@ -73,31 +73,15 @@ typedef struct _hclib_future_t {
...
-} hclib_triggered_task_t;
+struct hclib_task_t;
 
 // We define a typedef in this unit for convenience
 typedef struct hclib_promise_st {
     hclib_future_t future;
-	int kind;
-    volatile void *datum;
+    promise_kind_t kind;
+    void *volatile datum;
```
agrippa, line 82: Does it make sense for the pointer itself to be volatile? It seems like its value will only ever be set once and never changed again, what's the case where having the pointer storage itself be volatile is necessary?

inc/hcupc-support.h
```diff
@@ -103,6 +103,17 @@ inline void execute_hcupc_lambda(T* lambda) {
 			 * If a lambda is heap allocated then the pointer will not be valid at
 			 * remote place. Due to this we have to first copy the content of this
 			 * lambda into a stack allocated lambda.
+			 *
```
agrippa, line 106: Kumar will have to comment on this, I'm not very familiar with the HUPC code.

inc/hclib-promise.h
DaoWen, line 82: I'm pretty sure the original author thought that `volatile void*` was the equivalent of a volatile reference in Java—but it's not. (The HC runtime was designed based on the HJ runtime, so it wouldn't surprise me if someone was just translating a Java class into a C struct and did this.)
Since these fields in the promise struct are what get modified concurrently (not the values they point to), I think the intent was to mark the pointers themselves `volatile`.
However, as I pointed out in #18, I think we should be able to rid our codebase of `volatile` altogether. Using `__sync`/`__atomic` builtins should already be providing the memory barriers that we need to avoid stale values, so I don't think `volatile` is doing anything useful...
I left these `volatile`s for now because I'm only _pretty sure_ that we don't need them—but I'm 100% sure that they don't make any sense on the other side of the `*`.

inc/hclib-task.h
```diff
@@ -32,32 +32,20 @@ typedef struct _hclib_task_t {
      * and locality flexible asyncAny
      */
     int is_async_any_type;
```
vivkumar, line 34: "is_async_any_type" was to support asyncAny interface in HCUPC++. But I am pretty sure it won't work now (same reason that hclib codebase has changed a lot). Once resource worker is in master then we may think of a new design to support asyncAny

inc/hcupc-support.h
```diff
@@ -103,6 +103,17 @@ inline void execute_hcupc_lambda(T* lambda) {
...
+			 * so I don't see how this could be true. To further
+			 * complicate this, we're currently using heap-allocated
+			 * stacks for our fibers, so everything is in the heap
+			 * (i.e., even address-range-checking wouldn't work for
+			 * differentiating between stack and heap). This will
+			 * need to be refactored later, but I need to find out
+			 * what exactly it's supposed to be doing first.
 			 */
 
 			/*
```
vivkumar, line 119: Ok, let me try to think.. When a task (lambda) is popped from deque its a void\* pointer. The asyncAny interface specifies a task that can be either executed by a local worker or a remote worker (distributed work-stealing). In case of local execution, the above comment doesn't make sense. However, if a remote worker has to steal this task (which is a void\* pointer on the deque of remote computation worker), it can only be done if the type of lambda can be derived after the task is popped from the deque. We have to pass the content of lambda and not the pointer to this lambda at remote place. Thats why I was saving the "type" of lambda in the task structure which would help in doing memcpy of this lambda into a stack allocated lambda whenever a distributed work-stealing is happening.

However, now as entire support for HCUPC++ is broken, I am not sure if we really need this now.

DaoWen, line 119: It sounds like what was really needed here is a static assertion that the function object in question [`is_trivially_copyable`](http://en.cppreference.com/w/cpp/types/is_trivially_copyable#Notes). We might want to add such an assertion in the resource-worker-based HCUPC++ code in the future.
But yes, simply removing all of the HCUPC-specific code for now is probably easiest. (As you mentioned in the email, HCUPC++ users can instead use an older snapshot of HClib that is known to be compatible.)

inc/hclib-task.h
DaoWen, line 34: You might want to open an issue for that so that we don't forget.

src/hclib-promise.c
```diff
@@ -110,10 +100,8 @@ hclib_promise_t **hclib_promise_create_n(size_t nb_promises,
  * Note: this is concurrent with the 'put' operation.
  */
 void *hclib_future_get(hclib_future_t *future) {
-    if (future->owner->datum == UNINITIALIZED_PROMISE_DATA_PTR) {
-        return NULL;
-    }
-    return (void *)future->owner->datum;
```
DaoWen, line 116: This is related to the `volatile void *datum` change. This cast (and the similar casts in `hclib_future_wait`) are necessary because `datum` is a pointer to _volatile data_. After correcting the declaration, these casts are no longer necessary. The fact that we always discarded the `volatile` qualifier before dereferencing the `datum` pointer was a strong indicator that the declaration was incorrect.

inc/hclib-task.h
agrippa, line 34: I'm trying to get this merged in now that I've realized each of Nick's diffs includes all of the preceding diffs. Can we delete this is_async_any_type now? It really does not belong in hclib.

DaoWen, line 34: Based off Kumar's comments, I think it would be fine to delete `is_async_any_type`. However, since I don't actually touch that field in these changes, I think that's a separate issue that we can address in a later patch. I don't think this is an issue that should be holding up this pull request.

DaoWen, line 34: > I'm trying to get this merged in now that I've realized each of Nick's diffs includes all of the preceding diffs.

The changes build off of each other. As long as we go through the pull requests in order, each should be a manageable set of just a few commits. I thought that was clear from the last line in each of my pull request comments. The changes are all intertwined (later changes to promises affect the async_future function signatures, which were also updated in the earlier patches). Processing these pull requests in order (vs trying to split them into independent changes) should save us from some very hairy merges later.

agrippa, line 34: So then why open multiple dependent pull requests to begin with, if only one can be worked on at a time?

DaoWen, line 34: I forgot that GitHub sticks a stupid Merge Commit in the history after every PR is merged. Without that, the diffs would have shrunk on their own as we worked through them in order... Due to the merge commits, I'll have to rebase each successive PR as we move along, but that's still not too bad. Like I said, I thought the PR comment made it very clear that these changes are cumulative, and so you'd have no motivation to look at them out of order. I guess I should have prefixed the PR names with something like "[PART 1]", "[PART 2]" etc. to make this more obvious.... But you guys didn't have any problem getting through the first 3 PRs (and it seemed like you both understood that _those_ were cumulative), so I didn't have any reason to think that you guys would stall on the remainder of the chain.

agrippa, line 34: I still don't see an answer to my question. Why weren't they opened one at a time as they became actually mergeable diffs?

And we "stalled" because other things naturally arose Nick. I'm getting back to them now, but these aren't at the top of your or my TODO list.

Issue #17, opened 2016-06-22
Excessive copies of lambdas


DaoWen:
We need to refactor the C++ API to avoid passing function objects by value.

Currently, functions like `async` accept their function object arguments _by value_, which results in a lot of extra copies. For a typical `async` (see test [copies0](https://github.com/habanero-rice/hclib/blob/a86cc8d037182c0c29c6ff72f1bd57b47bf87c0c/test/cpp/copies0.cpp)), there are two or three extra copies. When using the `forasync` family of functions, the issue is much worse, with thousands of unneeded copies for a single 3D loop in from of our tests (see test [copies1](https://github.com/habanero-rice/hclib/blob/a86cc8d037182c0c29c6ff72f1bd57b47bf87c0c/test/cpp/copies1.cpp)).

PR #19, opened 2016-06-22
C++ async and forasync delegate to C API

DaoWen:
My understanding was that we wanted the C++ API to be a thin wrapper over the C API. These changes update the C++ API for async, forasync, etc. so that most of the internal logic is delegated to the C code.

This patch should also close issue #17, since it eliminates most of the extra copies of user functors that were happening in the C++ code in the async implementation, and more especially in forasync implementation.

---

<i>Note: This pull request builds on #16. You can compare the branches corresponding to the pull requests to see the diff between just #16 and this PR: https://github.com/habanero-rice/hclib/compare/cxx-cleanup-4...cxx-cleanup-5</i>


DaoWen:
Now that #16 is merged, I've rebased this PR off master.

DaoWen:
Superseded by #21.

PR #20, opened 2016-06-22
First pass on memory leaks

DaoWen:
Fixing a few of the memory leaks that were identified by FIXME comments in earlier commits.

---

<i>Note: This pull request builds on #19. You can compare the branches corresponding to the pull requests to see the diff between just #19 and this PR: https://github.com/habanero-rice/hclib/compare/cxx-cleanup-5...cxx-cleanup-6</i>

PR #21, opened 2016-06-22
Templatize promise and future types

DaoWen:
Template types with multiple specializations for both `promise_t` and `future_t`.

No more need for casts to/from `void*`! 😄 

Also generalized the async_future functions in the C++ API to allow the lambda to return a value (which is used as the corresponding future's value).

---

<i>Note: This pull request builds on #20. You can compare the branches corresponding to the pull requests to see the diff between just #20 and this PR: https://github.com/habanero-rice/hclib/compare/cxx-cleanup-6...cxx-cleanup-7</i>


inc/hclib-async.h
```diff
@@ -55,37 +56,69 @@ namespace hclib {
...
-}
 
-#if defined(HUPCPP) && defined(DIST_WS)	// i.e. if we are supporting distributed work-stealing in HabaneroUPC++
-#define _allocate_async _allocate_async_hcupc
-#else
-#define _allocate_async _allocate_async_hclib
-#endif
+/* this version also deletes a runtime-managed future list */
+template<typename T>
+struct LambdaAwaitArgs {
```
agrippa: Most of the types in HClib use underscores (including ones only referenced from the C++ side). Even if you prefer camel case for C++, let's try to remain consistent. I'm open to either option as long as it's done consistently.

inc/hclib-forasync.h
```diff
@@ -62,539 +62,171 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-		high1 = mid;
-		new_loop_initialized = true;
-	}
-	// recurse
-	if(new_loop_initialized) {
-		// delegate scheduling to the underlying runtime
-        auto lambda_wrapper = [=]() {
-            forasync2D_recursive<T>(new_loop, lambda, place, future_list);
-        };
+union ForAsyncFpUnion {
```
agrippa: Same type comment here.

inc/hclib.h
```diff
@@ -174,6 +175,14 @@ void hclib_end_finish();
 hclib_future_t *hclib_end_finish_nonblocking();
 void hclib_end_finish_nonblocking_helper(hclib_promise_t *event);
 
+
+#define _HCLIB_FINISH_IMPL(END_FUNC) \
```
agrippa, line 179: These don't seem to be used anywhere? What's the intended use case here?

src/hclib-runtime.c
```diff
@@ -354,14 +355,15 @@ static inline void check_out_finish(finish_t *finish) {
         // hc_atomic_dec returns true when finish->counter goes to zero
         if (hc_atomic_dec(&(finish->counter))) {
 #if HCLIB_LITECTX_STRATEGY
```
agrippa, line 357: What says everyone to making it so HCLIB_LITECTX_STRATEGY is the only option now? I doubt the other code path is well-tested at this point anyway.

agrippa:
I had only a few comments. Obviously this diff is massive, so it'll be very easy for me to miss things. Assuming all the tests are passing, I'd say it's ready to merge into master now.

inc/hclib-async.h
DaoWen: This change is part of #20. That's a good point though. I do prefer PascalCase for type names (in both C and C++), since that makes it easy to avoid name conflicts between types and data members; however, consistency obviously trumps my personal preferences, so I'll go back and rebase this to use snake_case instead.

inc/hclib-forasync.h
DaoWen: This change is part of #19. I actually want to get rid of this union altogether. If we split the C interface  to use 3 different functions for 1D, 2D, and 3D (which is how we have it in C++) then that would let us get rid of this nasty hack for converting function pointers to/from `void*`.

test/cpp/promise/future0Int.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include "hclib_cpp.h"
+
+
+int main(int argc, char ** argv) {
+    hclib::launch([]() {
+        constexpr int n_asyncs = 5;
+        hclib::future_t<int> *prev = nullptr;
+
+        HCLIB_FINISH {
```
DaoWen: Used `HCLIB_FINISH` here.

DaoWen: Used `HCLIB_FINISH` here.

inc/hclib.h
DaoWen, line 179: I used this macro in at least one of the tests. It's nice to have in C code since it makes finish scopes more obvious and helps ensure that the start/end calls always match. It's also nice in C++ since it lets you avoid unnecessarily introducing another lambda with another capture list vs the `hclib::finish` function.

src/hclib-runtime.c
DaoWen, line 357: I think we need to leave the alternatives. I have some other strategies that I will commit in the future, each with its own tradeoffs. However, I think having `HCLIB_LITECTX_STRATEGY` as the new default is a good thing.

DaoWen:
This diff actually isn't that big (it's only 3 commits). The problem is that it depends on #16, #19 and #20. We need to merge those first.

test/cpp/promise/future0Int.cpp
DaoWen, line 20: Used `HCLIB_FINISH` here.

src/hclib-runtime.c
vivkumar, line 357: Last time when I tried "not using" the HCLIB_LITECTX_STRATEGY, small changes were required in the runtime to fix the alternate code path.

vivkumar:
Hi Nick, I agree with Max reviews. Now if all the regression tests are passing with these changes, please go ahead with the merge.

DaoWen:
After a discussion with Zoran (and both reviewers here giving the go-ahead), I'm going ahead and merging this PR directly. I'll close #19 and #20.

PR #23, opened 2016-08-14
Simplify install process

DaoWen:
The installation process has been simplified. We now use the `xml2-config` script to get the correct compile flags for libxml2. Simply running `./install.sh` should be sufficient on most platforms for a basic local installation. The install script now creates an env setup script as part of the install process, which the user can source to correctly set `HCLIB_ROOT`. The installation instructions in the readme have also been updated accordingly.

The `scripts` directory has been removed since the `scripts/setup.sh` template is no longer necessary. If we want to keep access to the other setup scripts, maybe we should think about migrating them to another repository. Most of the non-obsolete settings are highly specialized, e.g., HCUPC and CUDA settings—neither of which are currently supported on the master branch.

---

<i>Note: This pull request builds on #21. You can compare the branches corresponding to the pull requests to see the diff between just #21 and this PR: https://github.com/habanero-rice/hclib/compare/cxx-cleanup-7...cxx-cleanup-8</i>


agrippa:
Uhhhh this seems like a commit that probably doesn't depend on all of your code changes. Can you create it in a branch off of master so that it isn't blocked on everything else? The diff is massive.

DaoWen:
This doesn't strictly depend on the other changes, but I think we should focus on merging the other changes (which have been sitting around since June) before we worry about cosmetic changes like simplifying the installation experience. I put this here intentionally. The diff should reduce to just the two relevant commits once the backlog is cleared.

agrippa:
So, I would argue that a large reason these changes have been lying around since June is that each of them looks massive. They aren't massive, but because you made them on top of each other reviewers are demotivated to look at these massive blobs as soon as they open them. Having them be so massive reduces the effectiveness of the review itself, as reviewer fatigue sets in. It also makes it impossible for reviewers to see what should be reviewed as part of that diff, and what is part of other diffs. It's good practice to make changes independently off of master wherever possible, and to hold off on putting out dependent reviews when it isn't possible. I would encourage you to make this diff (and any future ones) smaller and off of master if you want to encourage rapid and effective review of your changes:

https://blog.fogcreek.com/effective-code-reviews-9-tips-from-a-converted-skeptic/
https://smartbear.com/learn/code-review/best-practices-for-peer-code-review/

DaoWen:
> I would encourage you to make this diff (and any future ones) smaller and off of master if you want to encourage rapid and effective review of your changes.

Like I said, I want you guys to review the important, breaking API changes first. This was meant to be another friendly reminder to start working through the [pull request stack](http://graysonkoonce.com/stacked-pull-requests-keeping-github-diffs-small/). The fact that other people in the Habanero group are trying to install and use HClib makes it all the more urgent to get the breaking API changes merged. Kumar was busy with a couple of paper submissions last month, and I thought you were just busy with your projects at Waseda. I had no idea that you were trying to review the pull requests out of order until you made those unexpected comments on #21 yesterday.

agrippa:
I did the pull requests out of order because I lost context. Pull requests that were not intended to be reviewed yet should never have been opened in the first place. Can you see how that would be confusing for someone who does not have the same state in their head as you do in yours?

If people outside the Habanero group were using HClib, sure, API instability would be a problem. For people inside the Habanero group, no that is not a problem. There have been multiple people inside the group working on this shared repo for years now, it has not been a problem in the past and nothing has changed now.

Making this pull request dependent on the others only increases the latency between useful changes being made and merged. Who does that benefit?

DaoWen:
Closing in favor of #25, which contains the same two commits.

PR #28, opened 2016-11-28
Restore finish scope in hclib_future_wait

### LINKED ISSUE #27: Future wait operation does not restore finish scope
The future `wait` operation currently does not maintain the thread-local state of the current task. Specifically, the `current_finish` property is not restored after a potential fiber migration in blocking future `wait` calls. This can result in the wrong finish scope counter being decremented, and often triggers a segfault later in the execution due to broken synchronization.

The following test program demonstates this behavior:

https://github.com/habanero-rice/hclib/blob/8f649248e5a5d1a0e26dfa3097cce7fec31533db/test/cpp/promise/future4.cpp

DaoWen:
Closes #27.

### DIFF
--- a/src/hclib-runtime.c
+++ b/src/hclib-runtime.c
@@ -943,6 +943,9 @@ void *hclib_future_wait(hclib_future_t *future) {
         return future->owner->datum;
     }
 
+    // save current finish scope (in case of worker swap)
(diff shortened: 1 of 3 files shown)


agrippa:
Good catch. Interesting that we haven't had a test catch this yet.

PR #29, opened 2017-01-24
Calculate mid using non-overlow technique

mfrw:
Although, the values for which mid is calculated would probably never
overflow an int, still change the calculation method, just in case it
might.

### DIFF
--- a/src/hclib.c
+++ b/src/hclib.c
@@ -177,7 +177,7 @@ void forasync1D_recursive(void *forasync_arg) {
 
     //split the range into two, spawn a new task for the first half and recurse on the rest
     if((high0-low0) > tile0) {
-        int mid = (high0+low0)/2;
+        int mid = low0 + (high0-low0)/2;
         // upper-half
         forasync1D_task_t *new_forasync_task = allocate_forasync1D_task();
         new_forasync_task->forasync_task._fp = forasync1D_recursive;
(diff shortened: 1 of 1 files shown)


agrippa:
Looks great!

budimlic:
We should postpone merging this until we resolve the licensing issues and have a contributor agreement in place. Useful change but doesn't seem super-urgent.

After we switch to Apache 2.0 then we can merge this and it will be clear that the result is still under Apache 2.0 license

DaoWen:
@mfrw - Thanks for submitting this patch. Sorry it has taken so long to get to this. I looked at this back in January when you submitted it, but at the time we didn't have any policy in place for accepting contributions from outside the university. We decided to switch to the Apache 2.0 license to address this problem. Contributions from third parties (like this patch) are now explicitly covered by [clause 5 of our new license](https://github.com/habanero-rice/hclib/blob/6ad778a/LICENSE#L131-L137
):

> 5. Submission of Contributions. Unless You explicitly state otherwise, any Contribution intentionally submitted for inclusion in the Work by You to the Licensor shall be under the terms and conditions of this License, without any additional terms or conditions.  Notwithstanding the above, nothing herein shall supersede or modify the terms of any separate license agreement you may have executed with Licensor regarding such Contributions.

If the terms of the Apache 2.0 license are acceptable to you, please open another pull request with the same change, but rebased onto the current head revision. Thanks again for your contribution!

PR #32, opened 2017-03-31
Feature/C11 atomics

DaoWen:
Use the standard, portable C11 stdatomic.h for the atomic operations internal to HClib rather than the platform-specific assembly code in our current version. This was tested against a subset of [Max's tasking micro benchmarks](https://github.com/habanero-rice/tasking-micro-benchmark-suite), and there was no measurable difference in performance on x86_64 (guad.cs.rice.edu) or Power7 (biou.rice.edu).

### DIFF
--- a/src/hclib-deque.c
+++ b/src/hclib-deque.c
@@ -39,30 +39,36 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "hclib-internal.h"
 #include "hclib-atomics.h"
 
-void deque_init(deque_t *deq, void *init_value) {
+#if 0 // UNUSED
+void deque_init(deque_t *deq) {
     deq->head = 0;
+    //@ FIXME ATOMIC: this should be a RELEASE to ensure synchronization?
     deq->tail = 0;
 }
+
+void deque_destroy(deque_t *deq) {
+    free(deq);
+}
+#endif
 
 /*
  * push an entry onto the tail of the deque
  */
-int deque_push(deque_t *deq, void *entry) {
-    int size = deq->tail - deq->head;
+int deque_push(deque_t *deq, hclib_task_t *entry) {
+    int tail = _hclib_atomic_load_relaxed(&deq->tail);
+    int head = _hclib_atomic_load_relaxed(&deq->head);
+    int size = tail - head;
     if (size == INIT_DEQUE_CAPACITY) { /* deque looks full */
         /* may not grow the deque if some interleaving steal occur */
         // std::cout<<getenv("PMI_RANK") <<": Deque full for worker-"<<current_ws()->id << std::endl;
         // HASSERT("DEQUE full, increase deque's size " && 0);
         return 0;
     }
-    int n = (deq->tail) % INIT_DEQUE_CAPACITY;
-    deq->data[n] = (hclib_task_t *) entry;
-    deq->tail++;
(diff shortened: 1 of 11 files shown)


agrippa:
There appear to be segfaults in the testing. We'll need those triaged before I can do a review.

DaoWen:
The failures were due to [a bug with `__sync_bool_compare_and_swap` in clang 3.5](https://bugs.llvm.org//show_bug.cgi?format=multiple&id=21499). Switching to `__sync_val_compare_and_swap` fixes the problem that was causing those failures when using the legacy compiler support code.

src/hclib-deque.c
```diff
@@ -39,32 +39,38 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "hclib-internal.h"
 #include "hclib-atomics.h"
 
-void deque_init(deque_t *deq, void *init_value) {
+#if 0 // UNUSED
```
agrippa: If we're not using this code, why not just delete it? We have git if we really want to remember it at some point.

src/hclib-deque.c
```diff
@@ -97,39 +105,38 @@ hclib_task_t *deque_steal(deque_t *deq) {
...
+
     return t;
 }
 
 /******************************************************/
 /* Semi Concurrent DEQUE                              */
 /******************************************************/
 
-void semi_conc_deque_init(semi_conc_deque_t *semiDeq, void *initValue) {
+#if 0 // UNUSED
```
agrippa: I'd suggest deleting this too. This was only used by Kumar's HUPC++, and the support in the master branch doesn't work anymore anyway (as far as I'm aware). We have support in a branch, but that support also doesn't rely on semi_conc_deque.

agrippa (commented):
Looks good to me. Have you by any chance tried a few repeats of the C and C++ tests on a POWER machine? Most of this stuff looks like it wouldn't be much of a change on x86, but it might be on POWER.

src/hclib-deque.c
DaoWen: This 2-line change kept my diff smaller—but I agree we should delete this, as well as all the other unused/disabled code. There is *a lot* of it.

DaoWen:
> Have you by any chance tried a few repeats of the C and C++ tests on a POWER machine?

Yes, I tested this patch with both the `stdatomic.h` and legacy `__sync_*` configurations on BioU (applied on a different branch with powerpc64 support).

src/hclib-deque.c
agrippa: Okay, then I'd vote for now that we delete this block and the other block I commented on, and then I'll click the merge button.

DaoWen: I'd prefer actually removing all of the dead code in a single patch (or at least a set of consecutive commits), which should make it easier to track down if we need to find it again later.

DaoWen: Actually, removing this shouldn't be a problem since it's all limited to the deque. The other code I was thinking of is marked with `HC_CUDA`/`HCUPC` preprocessor conditionals and sprinkled all over the codebase. I'll delete this disabled code and update the branch shortly.

agrippa: Absolutely. I would be more than happy to see all of those CUDA and UPC++ blocks deleted and that logic removed from the configure scripts. I've already done this in my branch (and replaced them with my own support that doesn't rely on preprocessor macros).

PR #34, opened 2017-04-06
Remove obsolete/broken script and test code

DaoWen:
Remove several test files and scripts that either use missing/unsupported features.


agrippa (commented):
Looks good, just one quick question: since you've deleted all of the CUDA tests, do you have a separate pull request planned to delete the CUDA code in core?

DaoWen:
> do you have a separate pull request planned to delete the CUDA code in core?

Yes, I want to remove the HC_CUDA and HCUPC `#if-`guarded code—but I might not get around to it until next month.

PR #35, opened 2017-04-06
Update Cholesky decomposition example

DaoWen:
Fixed some errors to build it with the updated library.
Change malloc/free to new/delete since this is C++ code.
Switched from hclib::finish to more-efficient HCLIB_FINISH macro.


test/cholesky/cholesky.cpp
```diff
@@ -85,12 +83,12 @@ void entrypoint(void *arg) {
...
             TileBlock *currPivotColumnTile = lkji[j][k][k+1];
-            currPivotColumnTile->matrixBlock = (double**) malloc (sizeof(double*)*tileSize);
+            currPivotColumnTile->matrixBlock = new double*[tileSize];
             for(i = 0; i < tileSize; ++i)
-                currPivotColumnTile->matrixBlock[i] = (double*) malloc (sizeof(double)*tileSize);
+                currPivotColumnTile->matrixBlock[i] = new double[tileSize];
         }
 
-        finish ([=]() {
+        HCLIB_FINISH {
```
agrippa, line 91: This is more efficient just in that it skips a lambda creation?

DaoWen: Yes, it avoid creating an additional lambda/closure. Besides the efficiency issue, it's also nice from a maintenance perspective because you don't have to explicitly list captures or worry about unexpected captures due to a catch-all.

agrippa:
Looks like a compile-time error for gcc in the tests?

PR #40, opened 2017-04-06
Add -std=c11 flag to default project CFLAGS

DaoWen:
Needed for variadic macros and other more modern C features.
Tests using `HCLIB_FINISH` don't compile on legacy compilers without this.
Newer compilers have these features enabled by default,
but legacy compilers (e.g., gcc 4.8) need an explicit flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Dependencies
 
 * automake
 * gcc >= 4.8.4, or clang >= 3.5
-  (must support -std=c++11)
+  (must support -std=c11 and -std=c++11)
 * libxml2 (with development headers)
 
 
--- a/inc/hclib-mak/hclib.mak
+++ b/inc/hclib-mak/hclib.mak
@@ -2,10 +2,11 @@ ifeq ("$(HCLIB_ROOT)", "")
   $(error Please set teh HCLIB_ROOT environment variable.)
 endif
 
-PROJECT_CFLAGS=-I$(HCLIB_ROOT)/include $(shell xml2-config --cflags)
-PROJECT_CXXFLAGS=-std=c++11 $(PROJECT_CFLAGS)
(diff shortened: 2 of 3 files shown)


agrippa:
Why do we need this now?

DaoWen:
Our C tests' makefile depends on `PROJECT_CFLAGS`, which doesn't include `-std=c11`. This means that if any C test is updated to use the `HCLIB_FINISH`, it won't compile because the `HCLIB_FINISH` macro implementation depends on post-C89 features.

agrippa:
Wouldn't -std=c99 be sufficient? What in c11 do we require?

DaoWen:
We already require features to C11 to build the runtime. C99 would be sufficient to get `HCLIB_FINISH` to work, but it seems silly to explicitly specify a standard which is less featureful than the one we require to build the runtime. In the future, it might be a good idea to conditionally generate `PROJECT_CFLAGS` during the installation so that we don't override the standard if the default is a post-C11 standard—but I'm pretty sure the current versions of GCC and Clang just use `gnu11` by default, so this should be fine for now.

agrippa:
What c11 features do we already require?

DaoWen:
See 1342ab68fc for one example.

agrippa:
Then should we change the macro you linked to so that it #errors out when c11 support isn't present? It sounds like you're saying c11 support should be required, but that macro is inconsistent?

DaoWen:
I should go back and simplify those assertion macros. Anyway, this patch is only fixing an inconsistency between our runtime build flags and our test build flags. Our runtime build system already assumes C11, and the compilers listed in our README all support C11. I don't see any compelling reason to revert back to an earlier version of C.

agrippa:
Sure, so then my question was "should we change the macro you linked to so that it #errors out when c11 support isn't present?" This would seem to address the same type of inconsistency.

PR #42, opened 2017-04-07
Simplify HASSERT_STATIC definition

DaoWen:
See discussion in #40.
Since we require C11 and C++11 in the build script,
we can safely assume that static assertions are available.

### DIFF
--- a/inc/hclib-rt.h
+++ b/inc/hclib-rt.h
@@ -93,15 +93,10 @@ typedef struct hclib_worker_state {
 #define HASSERT(cond)       // Do Nothing
 #endif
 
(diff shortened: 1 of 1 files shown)


agrippa:
Looks good, could you just change the following in README.md:

"must support -std=c++11"

to be:

"must support -std=c11 and -std=c++11"

or something to that effect.

DaoWen:
I'm planning to update the README later to clarify some installation details. I'll update the dependencies then as well.

agrippa:
This seems closely related to this change, whereas some other dependencies are not. Why not now?

DaoWen:
I added the README update to #40 because it seems more relevant there.

PR #43, opened 2017-04-07
Update deadlock test and add to default C tests

DaoWen:
Move the deadlock proof-of-concept into our C test suite. Also swap its local `FINISH` macro definition with the `HCLIB_FINISH` now provided by the library.


agrippa:
This looks good. It does raise the question: do we rip out the non-fiber code path now? (either way, the test is good to have). Any opinions?

PR #46, opened 2017-04-13
Create CONTRIBUTING.md

budimlic:
Adding a file that describes the procedure and guidelines for creating pull requests. A link to this file will be automatically generated by GitHub and shown to anyone creating a pull request.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,205 @@
+# Contributing to HClib
+
+Before opening a Pull Request,
+please be sure to read through these guidelines.
+Following the guidelines listed below will make it much more likely
+for your contributions to be quickly accepted and merged.
+
+## Licensing
+
+All contributions submitted to this project are governed
+by the terms outlined in our license agreement.
+Specifically, our license (Apache 2.0)
+has an explicit contributions clause in [section
+5](https://github.com/habanero-rice/hclib/blob/6ad778a/LICENSE#L131-L137):
+
+> Unless You explicitly state otherwise, any Contribution intentionally
+> submitted for inclusion in the Work by You to the Licensor shall be under
+> the terms and conditions of this License, without any additional terms or
+> conditions.  Notwithstanding the above, nothing herein shall supersede or
+> modify the terms of any separate license agreement you may have executed
+> with Licensor regarding such Contributions.
+
+## Pull Requests
+
+Please follow established best practices for your pull requests.
+Give a short description of *what* your Pull Request does in the title,
+and include a short description of *why* the Pull Request is needed in the comment.
+A complex title or summary is a good indication that your Pull Request
+is doing too many things, and should be split into multiple Pull Requests.
+In general, your Pull Request should be doing a single thing whenever possible.
+
+Your Pull Request should be squashed to a single commit, in the spirit of making small,
+incremental, and separable changes, and to simplify the log. Please do not squash your unrelated commits just to 
+comply with this request, if you do so you will be asked to separate your changes. Your commit message should
+explain *what* and *why* your commit does, not *how*. The *how* should be explained in your code comments.
+
+When you create a Pull Request, please assign one or more 
+[`core team members`](https://github.com/orgs/habanero-rice/teams/habanero-rice-devs) 
+as the reviewers. 
+
+Once you create a Pull Request, GitHub will automatically start a discussion 
+concerning your request. You should monitor the discussion and address the suggestions,
+concerns, and critiques from other collaborators.
+
+### Merging Pull Requests
+
+Only the 
+[`Core team members`](https://github.com/orgs/habanero-rice/teams/habanero-rice-devs)
+have merge privileges for Pull Requests in the master branch, but all collaborators
+are welcome to review and discuss a particular Pull Request. 
+
+Every Pull Request must be signed off by two Core team members in order to be merged. After one Core member completes their review and is satisfied that all their comments are addressed they should leave a comment indicating their approval. After a second Core member completes a review and had all comments addressed, they may perform the actual merge.
+
+Core members should **never** merge their own Pull Requests, they must follow the same guidelines above as all the other contributors and have their Pull Request reviewed and merged by two other Core members.
+
+
+## Code Review Guidelines
+Here's a nice set of 
+guidelines to follow when reviewing and discussing contributor's code (borrowed heavily and modified from 
+[`Thoughtbot`](https://github.com/thoughtbot/guides/tree/master/code-review)
+): 
+
+### Everyone
+
+* Accept that many programming decisions are opinions. Discuss tradeoffs, which
+  you prefer, and reach a resolution quickly.
+* Ask good questions; don't make demands. ("What do you think about naming this
+  `:user_id`?")
+* Avoid judgment and avoid assumptions about the author's
+  perspective.
+* Ask for clarification. ("I didn't understand. Can you clarify?")
+* Avoid selective ownership of code. ("mine", "not mine", "yours")
+* Avoid using terms that could be seen as referring to personal traits. ("dumb",
+  "stupid"). Assume everyone is intelligent and well-meaning.
+* Be explicit. Remember people don't always understand your intentions online.
+* Be humble. ("I'm not sure - let's look it up.")
+* Don't use hyperbole. ("always", "never", "endlessly", "nothing")
+* Don't use sarcasm.
+* Keep it real. If emoji, animated gifs, or humor aren't you, don't force them.
+  If they are, use them with aplomb.
+* Talk synchronously (e.g. chat, screensharing, in person) if there are too many
+  "I didn't understand" or "Alternative solution:" comments. Post a follow-up
+  comment summarizing the discussion.
+
+### Having Your Code Reviewed
+
+* Be grateful for the reviewer's suggestions. ("Good call. I'll make that
+  change.")
+* Don't take it personally. The review is of the code, not you. Be aware of how hard it is to convey 
+  emotion online and how easy it is to misinterpret feedback. 
+* Keping the previous point in mind: assume the best intention from the reviewer's comments.
+* Explain why the code exists. ("It's like that because of these reasons. Would
+  it be more clear if I rename this class/file/method/variable?")
+* Seek to understand the reviewer's perspective.
+* Try to respond to every comment.
+
+### Reviewing Code
+
+Understand why the change is necessary (fixes a bug, improves the user
+experience, refactors the existing code). Then:
+
+* Communicate which ideas you feel strongly about and those you don't.
+* Identify ways to simplify the code while still solving the problem.
+* If discussions turn too philosophical or academic, move the discussion offline
+* Offer alternative implementations, but assume the author already considered
+  them. ("What do you think about using an off the shelf library here?")
+* Seek to understand the author's perspective.
+* Sign off on the Pull Request with a :thumbsup: or "Ready to merge" comment.
+
+## Quality Control for Contributed Code
+
+### Regression Tests
+
+The main regression tests for HClib are located in `test/c` and `test/cpp`.
+The `test_all.sh` scripts in each of those directories will automatically
+build and run all tests in the respective test suite.
+
+All new commits should pass all of the regression tests.
+When you open a Pull Request, Travis-CI will automatically run
+the full regression suite against your cumulative change
+(not against individual commits).
+Please run the tests locally against each commit to ensure
+that intermediate commits do not fail any tests
+(which breaks the ability to use `git bisect` and similar tools).
+
+If your Pull Request fixes a bug, it should also include a regression test for detecting that bug (with the appropriate changes in `test/c`, `test/cpp` and `test_all.sh`. This is to reduce the possibility of reintroducing the bug with some later changes.
+
+If your Pull Request implements a new feature, it should also include tests in `test/c` and `test/cpp` that test that feature.
+
+### Static Error Checking
+
+As part of the development workflow for HClib, any newly committed code
+should be checked using standard static checking tools.
+
+In particular, run
+[`cppcheck`](https://sourceforge.net/projects/cppcheck/)
+on all modified files.
+cppcheck should be run by `cd`-ing to `tools/cppcheck` and executing
+the `run.sh` script from there (this assumes `cppcheck` is on your path).
+Any new errors printed by `cppcheck` should be addressed before committing.
+
+### Code Formatting
+
+You should also run `astyle` on all modified or newly created files.
+`astyle` is a source code auto-formatter.
+Simply `cd` to `tools/astyle` and execute the `run.sh` script.
+This assumes you have `astyle` installed and it is on your path.
+
+## Summary
+
+Here are some useful tips for making good Pull Requests (borrowed and heavily modified from 
+[`Mark Seemann's blog`](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/)
+):
+
+1. *Make it small.*
+A small, focused Pull Request gives you the best chance of having it accepted. The time it takes to review a Pull Request 
+is typically non-linear in the size of the request. Don't make your reviewers work.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+the reviewers. You should also assign all other contributors who may be affected
+by your changes.
+
+Once you create a pull request, GitHub will automatically start a discussion 
+concerning your request. You should monitor the discussion and address the suggestions,
+concerns and critiques from other collaborators.
+
+### Merging Pull Requests
+
+Only the Core team members have the merge privileges for pull requests in the master branch, but all collaborators
```
DaoWen: Maybe we should concretely define "core team" here by creating a `habanero-rice/hclib-core` team and linking to it here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+are welcome to review and discuss a particular pull request. Here's a nice set of 
+guidelines to follow when reviewing and discussing contributor's code: 
+https://github.com/thoughtbot/guides/tree/master/code-review
+
+Once a consensus has been reached among the contributors, a core member can manually perform the merge. 
+
+Core members should **never** merge their own pull requests, unless in a high emergency situation (for example, a Core member
```
DaoWen: I think we should delete everything in this sentence after the comma. Once sensitive data is pushed to GitHub, it's compromised, so just reverting the change is insufficient. I can see how it might still be a good idea to do some damage control, but in that case you'd need to force-push directly to master, not open a pull request. Pull requests stay around forever (they're impossible to delete), so if you open a pull request to erase sensitive data, the sensitive data is still visible in the pull request. But yeah, I don't see a compelling reason to include this type of exception here.

DaoWen (changes requested):
This looks pretty good overall, but I think we should delete the exception about merging your own PRs.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+5](https://github.com/habanero-rice/hclib/blob/6ad778a/LICENSE#L131-L137):
+
+> Unless You explicitly state otherwise, any Contribution intentionally
+> submitted for inclusion in the Work by You to the Licensor shall be under
+> the terms and conditions of this License, without any additional terms or
+> conditions.  Notwithstanding the above, nothing herein shall supersede or
+> modify the terms of any separate license agreement you may have executed
+> with Licensor regarding such Contributions.
+
+## Commits
```
agrippa: The section on pull requests should be above the one on commits. I'm not even sure we need a section on commits, as there could just be a paragraph in the section on pull requests talking about commit messages. This guideline of "do only one thing per commit" should really be "do only one thing per pull request", so while it isn't wrong it also doesn't capture the whole spirit of this guideline.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+
+Please follow established best practices for your commits.
+Do only one thing per commit, and write a descriptive commit message.
+[Chris Beams' *seven rules of a great Git commit
+message*](https://chris.beams.io/posts/git-commit/#seven-rules)
+provides a good set of guidelines:
+
+1. Separate subject from body with a blank line
```
agrippa: In general, our commits do not have subject lines and a body. I think copy-pasting these guidelines goes against the existing state of the repo, and in general I disagree with several of them (or think they are unnecessarily restrictive). We'd have to have a discussion if we wanted all of our commit messages to start following these principles.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Pull Requests
+
+Please follow established best practices for your pull requests.
+Give a short description of *what* your patch does in the title,
+and include a short description of *why* the patch is needed in the comment.
+A complex title or summary is a good indication that your pull request
+is doing too many things, and should be split into multiple patches.
+
+See also [*10 tips for better Pull Requests* by Mark
```
agrippa: Rather than just pointing to external resources, I think we should heavily modify and include these guidelines inline. What if the external resource changes? What if the link breaks? What if some of the things they say don't really apply? For example, this linked resource says "Add tests". That's a pretty useless guideline if we don't also tell people how to add tests, where to put them, etc.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+Seemann](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/).
+
+When you create a pull request, please assign one or more core team members as
```
agrippa: How does an external committer know 1) who the core team is, and 2) other contributors who may be affected by your changes?

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+Once you create a pull request, GitHub will automatically start a discussion 
+concerning your request. You should monitor the discussion and address the suggestions,
+concerns and critiques from other collaborators.
+
+### Merging Pull Requests
+
+Only the Core team members have the merge privileges for pull requests in the master branch, but all collaborators
+are welcome to review and discuss a particular pull request. Here's a nice set of 
+guidelines to follow when reviewing and discussing contributor's code: 
+https://github.com/thoughtbot/guides/tree/master/code-review
```
agrippa: Again, external resources shouldn't be used to define core operating principles.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+
+Once a consensus has been reached among the contributors, a core member can manually perform the merge.
```
agrippa: How do the core members coordinate on consensus? How is it enforced that something isn't merged before everyone agrees? Again, how do we determine who a review is blocked on?

CONTRIBUTING.md
agrippa: Agreed. Pushing to master is a privilege, not a right. It should not be abused, nor should any changes to master (pushes, reverts, etc) be done unilaterally.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+https://github.com/thoughtbot/guides/tree/master/code-review
+
+Once a consensus has been reached among the contributors, a core member can manually perform the merge. 
+
+Core members should **never** merge their own pull requests, unless in a high emergency situation (for example, a Core member
+has mistakenly pushed some proprietary or security-sensitive data into the master and needs to immediately revert the change).
+
+## Quality Control for Contributed Code
+
+### Regression Tests
```
agrippa: Guidelines for including tests in new pull requests?

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+
+The main regression tests for HClib are located in `test/c` and `test/cpp`.
+The `test_all.sh` scripts in each of those directories will automatically
+build and run all tests in the respective test suite.
+
+All new commits should pass all of the regression tests.
+When you open a pull request, Travis-CI will automatically run
+the full regression suite against your cumulative change
+(not against individual commits).
+Please run the tests locally against each commit to ensure
```
agrippa: Should we ask for each pull request to be squashed to a single commit, in the spirit of making small, incremental, separable changes? Which would also make this advice unnecessary.

agrippa (commented):
What about guidelines for reviewers?

Guidelines for when things should be reverted vs. when a follow-on pull request should be requested?

CONTRIBUTING.md
budimlic: OK, so I'll delete the section on commits and adding a paragraph with commit guidelines to the section on pull request. I still think there should still be some commit message guidelines, I like 1., 5. and 7.

CONTRIBUTING.md
budimlic: OK, so I'll delete the section on commits and adding a paragraph with commit guidelines to the section on pull request. I still think there should still be some commit message guidelines, I like 1., 5. and 7.

CONTRIBUTING.md
budimlic: I removed Shams from @habanero-rice/habanero-rice-devs and added Nick. I think that should be the Core team (with perhaps moving Vivek out of it and leave him as an Owner). Everyone else is an Ouside collaborator.

CONTRIBUTING.md
budimlic: OK, agreed.

CONTRIBUTING.md
budimlic: Will link to @habanero-rice/habanero-rice-devs as the core team. Checking who previously committed the code they are modifying is a good indication to who might be interested/affected by what they are adding.

CONTRIBUTING.md
budimlic: Fair enough. I'll import and rewrite these guidelines.

CONTRIBUTING.md
budimlic: Well, that's a tough one. Asking for a formal vote/consensus of all core member does seem like an overkill. On the other hand, I do want to make sure everyone on the core team has had a chance to take a look at the pull request and comment. Suggestions are welcome here. Maybe a  time period after which any core member can perform a merge if all the core members that did comment on the pull request seem to be satisfied with the state of it?

CONTRIBUTING.md
budimlic: Sure. If people start squashing their unrelated commits just to comply with this request, we can always ask them to separate their changes when they submit the pull request

CONTRIBUTING.md
agrippa: Could you show me an example of the type of commit you're thinking of? For example, at the moment I don't believe a single commit in this repo has both a subject and a body, so that would be a pretty disruptive change to everyone's development practices today. That doesn't mean it's a bad idea, just something that's definitely going to need more discussion. In general, most repos don't have subject and body, but if it's something you feel strongly about we can definitely talk about adding that requirement.

CONTRIBUTING.md
agrippa: Okay, I agree with adding the first part to the text. But the second part seems like it should just be the responsibility of the "core team" reviewing the code to judge the disruptiveness of the change and decide if any special notification is required? We want to have a rigorous process but also not an onerous one, otherwise we de-incentivize contributions. Asking someone to figure out everyone who's committed to a file even in just the last month sounds painful to me, and is likely to turn others away.

CONTRIBUTING.md
agrippa: In past jobs, one convention I've worked with is that reviews are only integrated after 2 reviewers sign off, one of which is a "gatekeeper". A "gatekeeper" is basically an expert, in this case analogous to what we're calling a member of the "core team". Now, we don't really have non-gatekeepers at the moment; pretty much everyone pushing to this repo right now would qualify as one. So we could just say that two members of the core team are required to sign off on each review. After the first completes their review and all comments are addressed they simply leave a comment or something indicating their approval. Once the second has similarly completed a review and had all comments addressed, they do the actual merge.

The only complication for our present state is that Nick and I are the most active in terms of code changes and reviews (though most of my pushes aren't going into master), so to have a third reviewer who isn't the person putting out the review would require you (Zoran) to be responsive to regular review requests. However, if the review guidelines are followed then the result should be that those reviews don't consume large amounts of time.

CONTRIBUTING.md
agrippa: Agreed. Things shouldn't be squashed for convenience, just for simplifying the log.

CONTRIBUTING.md
```diff
@@ -0,0 +1,205 @@
...
+> submitted for inclusion in the Work by You to the Licensor shall be under
+> the terms and conditions of this License, without any additional terms or
+> conditions.  Notwithstanding the above, nothing herein shall supersede or
+> modify the terms of any separate license agreement you may have executed
+> with Licensor regarding such Contributions.
+
+## Pull Requests
+
+Please follow established best practices for your pull requests.
+Give a short description of *what* your patch does in the title,
```
agrippa: It looks like here we're using two words (pull requests + patches) to mean the same thing. Can we just always refer to it as a pull request?

agrippa: e.g. below it says "A complex title or summary is a good indication that your pull request is doing too many things, and should be split into multiple patches" when it seems like it would be simpler to just say "A complex title or summary is a good indication that your pull request is doing too many things, and should be split into multiple pull requests"

CONTRIBUTING.md
```diff
@@ -0,0 +1,205 @@
...
+comply with this request, if you do so you will be asked to separate your changes. Your commit message should
+explain *what* and *why* your commit does, not *how*. The *how* should be explained in your code comments.
+
+When you create a pull request, please assign one or more 
+[`core team members`](https://github.com/orgs/habanero-rice/teams/habanero-rice-devs) 
+as the reviewers. 
+
+Once you create a pull request, GitHub will automatically start a discussion 
+concerning your request. You should monitor the discussion and address the suggestions,
+concerns and critiques from other collaborators.
```
agrippa: oxford comma? "suggestions, concerns, and critiques"

CONTRIBUTING.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Merging Pull Requests
+
+Only the 
+[`Core team members`](https://github.com/orgs/habanero-rice/teams/habanero-rice-devs)
+have the merge privileges for pull requests in the master branch, but all collaborators
```
agrippa: "have merge privileges", delete "the"

CONTRIBUTING.md
```diff
@@ -0,0 +1,205 @@
...
+are welcome to review and discuss a particular pull request. 
+
+Every pull request must be signed off by two Core team members in order to be merged. After one Core member completes their review and is satisfied that all their comments are addressed they should leave a comment indicating their approval. After a second Core member completes a review and had all comments addressed, they perform the actual merge.
```
agrippa: "they may perform the actual merge", insert "may"

CONTRIBUTING.md
```diff
@@ -0,0 +1,205 @@
...
+This assumes you have `astyle` installed and it is on your path.
+
+## Summary
+
+Here are some useful tips for making good pull requests (borrowed and heavily modified from 
+[`Mark Seemann's blog`](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/)
+):
+
+1. *Make it small.*
+A small, focused Pull Request gives you the best chance of having it accepted. The time it takes to review a Pull Request
```
agrippa, line 156: Inconsistent capitalization (or not capitalization) of pull request. I'm fine with either capitalizing or not, just want it consistent.

CONTRIBUTING.md
```diff
@@ -0,0 +1,205 @@
...
+1. *Make it small.*
+A small, focused Pull Request gives you the best chance of having it accepted. The time it takes to review a Pull Request 
+is typically non-linear in the size of the request. Don't make your reviewers work.
+
+2. *Do only one thing.*
+Your Pull Request should address only a single concern. If it addresses multiple concerns (say A, B and C), the reviewers might 
+be fine with A and B, but have issues with C, delaying the merge of your Pull Request. This might unecessarily complicate the merge 
+of your perfectly fine A and B contributions if there were some updates to the repository in the meantime.
+
+3. *Watch your line width.*
```
agrippa, line 164: I would add that we encourage any newly merged code to be no wider than 80 columns.

agrippa (commented):
Looks great! I think with just a few small changes this'll be good to go and then Nick can merge it once he finishes his review. I'm sure we'll run into cases not covered by this doc, but we can tackle those issues when we come to them. Thanks for doing this!

DaoWen (approved):

agrippa:
This'll need to get squashed to one commit before merging (as stated in the new contributing guidelines!)

budimlic:
I'll squash all the changes into a single commit once Max approves the changes

agrippa:
Yep, looks good. Just let me know when it's ready and I'll merge it.

budimlic:
Done!

DaoWen:
👍  In the spirit of keeping the log simple, I'm going to do the "Squash and merge" operation rather than the default merge operation. That way we won't have so many pull request merge commits cluttering our commit log.

PR #47, opened 2017-04-14
Only set HAVE_C11_STDATOMIC if we have the _Atomic keyword

agrippa:
I found that the master build was broken on Edison because while icc 17.0.1 had stdatomic.h, it still didn't recognize _Atomic as a keyword. However, we were only checking for stdatomic.h before setting HAVE_C11_STDATOMIC.


agrippa:
Not to be reviewed/merged before #46

DaoWen:
I guess icc is using gcc's headers. This looks like a good workaround. 👍

agrippa:
@budimlic ?

budimlic:
Yep, this is a good workaround.

budimlic (approved):

agrippa:
@budimlic You can merge because you were the second to approve, per the contributor's agreement. I am not allowed to merge.

PR #48, opened 2017-04-14
Deadlock example was creating redundant promises

agrippa:
From what I can tell, promise objects were being created in the deadlock0 test which were not being used, leading to the code being confusing (correct me if I'm wrong). I've just made a small change to clean up that section of the code.

### DIFF
--- a/test/c/deadlock0.c
+++ b/test/c/deadlock0.c
@@ -18,6 +18,7 @@
 #include <stdio.h>
 #include <unistd.h>
 #include <stdbool.h>
+#include <assert.h>
 
 #include "hclib.h"
 
@@ -41,7 +42,6 @@ static inline void echo_worker(const char *taskName) {
     printf("Task %s run by worker %d\n", taskName, get_current_worker());
 }
 
-hclib_promise_t **promise_list;
 hclib_future_t **future_list;
 hclib_promise_t *promise;
 int data;
@@ -76,11 +76,12 @@ void taskC(void *args) {
 }
 
 void taskMain(void *args) {
-    promise_list = hclib_promise_create_n(1, true);
-    future_list = (hclib_future_t **)malloc(1 * sizeof(hclib_future_t *));
     promise = hclib_promise_create();
-    promise_list[0] = promise;
+
+    future_list = (hclib_future_t **)malloc(2 * sizeof(hclib_future_t *));
+    assert(future_list);
     future_list[0] = hclib_get_future_for_promise(promise);
+    future_list[1] = NULL;
 
     hclib_async(&taskA, NULL, NO_FUTURE, NO_PHASER, ANY_PLACE, NO_PROP);
     hclib_async(&taskC, NULL, NO_FUTURE, NO_PHASER, ANY_PLACE, NO_PROP);
--- a/src/hclib-promise.c
+++ b/src/hclib-promise.c
@@ -61,6 +61,8 @@ hclib_promise_t **hclib_promise_create_n(size_t nb_promises,
         int null_terminated) {
     hclib_promise_t **promises = (hclib_promise_t **) malloc((sizeof(
                                      hclib_promise_t *) * nb_promises));
+    HASSERT(promises);
+
     int i = 0;
     int lg = (null_terminated) ? nb_promises - 1 : nb_promises;
     while (i < lg) {


agrippa:
Not to be reviewed/merged before #46

budimlic:
This is true, the hclib_promise_create_n will create a promise and initialize promise_list[0] with it, then the test will later overwrite promise_list[0].

Why not use the hclib_promise_create_n call as it is right now, then just assign promise_list[0] to promise? Seems like the new code is duplicating the work done in hclib_promise_create_n

agrippa:
Mostly because I think that hclib_promise_create_n is a confusing API so I was trying to minimize its usage. For example, I just realized after looking at it closer that this code was technically not creating redundant promises. By calling:

hclib_promise_create_n(1, true);

where 1 is a parameter called nb_promises and true is passed as a parameter called null_terminated, you would expect that it would create an array of length 2 with the first slot holding a promise and the second being NULL. Indeed, your comment indicates that's how you read it too.

In fact, all it does is allocate an array of length 1 and initialize the 0th slot to NULL. Which is incredibly counter-intuitive. Therefore, I would argue this is an improvement in readability.

budimlic:
Yes, you are correct, hclib_promise_create_n(n,flag) creates an array of n promises, with the last one being NULL if the flag is set to true. That's really counterintuitive. But I think that's a topic for another pull request (create separate hclib_promise_create_n and hclib_promise_create_n_null_terminated).

I don't see a point in using an array of promises in this particular test, though. It's never used. Only the scalar promise variable is used in the test. So I suggest getting rid of the promise array completely.

test/c/deadlock0.c
```diff
@@ -76,11 +77,14 @@ void taskC(void *args) {
...
-    promise_list = hclib_promise_create_n(1, true);
+    promise_list = (hclib_promise_t **)malloc(2 * sizeof(hclib_promise_t *));
     future_list = (hclib_future_t **)malloc(1 * sizeof(hclib_future_t *));
-    promise = hclib_promise_create();
-    promise_list[0] = promise;
+    assert(promise_list && future_list);
+
+    promise_list[0] = hclib_promise_create();
+    promise_list[1] = NULL;
     future_list[0] = hclib_get_future_for_promise(promise);
```
budimlic: This looks like a bug. You're trying to extract a future from the promise variable, which is uninitialized at this point

DaoWen:
One reason I used `hclib_promise_create_n` in this test was because it's not called in any of our other tests. If it's in the API, then we should use it. However, I don't see why we even have an option for not using the NULL terminator.

I agree the API is confusing (hence my screw-up here). I'd prefer removing the boolean argument, and always allocating n+1 slots, with the last slot always set to NULL.

In addition to the bug that Zoran pointed out, there's another bug in this test, which is also in the unpatched version: the `future_list` isn't NULL-terminated. The future pointers are the arguments that are actually passed into `async_await`, and that's what really needs to be NULL-terminated.

We assume internally that promise and future pointers can be directly cast back and forth, which means this must also work:

    future_list = (hclib_future_t **) promise_list;

But that's a bit cryptic, so maybe we should add the following function to the C API:

    static inline hclib_future_t **hclib_get_futures_for_promises(hclib_promise_t **promises) {
        HASSERT_STATIC(offsetof(hclib_promise_t, future) == 0, "Can't cast promise ptr to future ptr");
        return (hclib_future_t **) promises;
    }

and use it like this:

    future_list = hclib_get_futures_for_promises(promise_list);

Since `promise_list` is NULL-terminated, the resulting `future_list` is also correctly NULL-terminated.

DaoWen:
👍 It looks like Max fixed the NULL terminator issue with his latest patch. I'm fine with merging this. We still need to figure out what we want to do with `hclib_promise_create_n`, but that's really a separate issue.

DaoWen (approved):
👍 It looks like Max fixed the NULL terminator issue along with the `promise` bug in the updated patch. I'm fine with merging this. We still need to figure out what we want to do with `hclib_promise_create_n`, but that's really a separate issue.

budimlic (approved):
Looks good now

agrippa:
Just continuing the discussion here.

I can see two solutions to the hclib_promise_create_n issue:

1. Simply change the behavior of hclib_promise_create_n to be more intuitive the way Nick suggested above. (always allocate n+1 slots, and always null terminate).
2. Change the await APIs to take the number of futures as an additional argument, rather than using null termination to implicitly indicate the number of futures. This would make hclib_promise_create_n less necessary.

I've actually implemented the second option in a branch because I always found the null-termination semantics to be a bit confusing, subtle, and error prone. I'm also okay with the first option, but I'd like to hear what you guys think.

DaoWen:
I'd prefer option 2 (the explicit count argument) because it's more flexible (e.g., you can await different sub-ranges of a single array in two different asyncs), and it would help internally in some places as well (e.g., we wouldn't need [this hack](https://github.com/habanero-rice/hclib/blob/ec848b9cc80e7026c51595ce11ce2f5f2e957189/src/hclib-runtime.c#L1148-L1160)). Obviously this will require some refactoring, but it's probably worth it.

budimlic:
That's a very good point Nick. I like option 2 as well.

PR #49, opened 2017-05-09
Fix pthread_setconcurrency warning

DaoWen:
Ensure X/Open 6 features are enabled to avoid compiler
warnings about undeclared functions on Linux.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -164,6 +164,9 @@ fi
 CXXFLAGS="${_old_CXXFLAGS}"
 ### End C++11 features check
 
+# We require at least POSIX.1-2001 for the pthread_setconcurrency function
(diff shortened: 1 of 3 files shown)


agrippa:
Have we ever actually seen pthread_setconcurrency improve performance? My understanding of it suggested that there aren't really any use cases where we'd expect it to help. What do you think? Should we just remove it?

This'll also be blocked until the merge guidelines can be integrated (like my two pull requests).

DaoWen:
The comment above the call suggested that it helps when you have oversubscription, but after reading the POSIX documentation I think that's inaccurate. It looks like the call is only useful in "N:M" threading systems—which seem to have all died out in the POSIX world. Let's just delete it. I'll open a new pull request.

PR #50, opened 2017-05-10
Remove obsolete pthread_setconcurrency call

DaoWen:
The function is a no-op in pthreads implementations on modern systems,
and including the call requires extra configuration to avoid compiler
warnings when building on Linux.


agrippa:
Nice. Looks good to me. @budimlic, you should take a look at this and merge it once Nick merges CONTRIBUTING.md in.

budimlic (approved):

PR #51, opened 2017-05-12
Simplify .gitignore for C tests

DaoWen:


DaoWen:
We weren't ignoring the `boot0` or `deadlock0` test binaries. The more general wildcards make it so we won't need to modify the .gitignore when adding new tests unless the prefix is totally new.

budimlic:
What happened to *.dSYM?

DaoWen:
> What happened to *.dSYM?

It's no longer needed since `async0.dSYM` is covered by the `async*` rule, etc.

budimlic:
Gotcha. I would only add another rule for not excluding .h files (which we currently don't have in the test directory, but might in the future)

test/c/.gitignore
```diff
@@ -1,20 +1,8 @@
...
-forasync2DRec
-forasync3DCh
-forasync3DRec
+async*
+boot*
+deadlock*
+future*
+finish*
+forasync*
+!*.[ch]
```
DaoWen, line 8: Now we exclude `*.c` and `*.h` files from being ignored.

agrippa:
Looks good to me. @budimlic can merge once he signs off.

PR #52, opened 2017-05-16
Remove disabled feature code

### LINKED ISSUE #24: Delete is_async_any_type field of struct _hclib_task_t in hclib-task.h

DaoWen:
Delete unused code that previously supported CUDA, UPC and SHMEM.

Closes #24.


inc/hclib-async.hpp
```diff
@@ -100,19 +99,19 @@ struct LambdaFutureWrapper<T, void> {
...
- * name to async_at_hpt.
- */
 template <typename T>
 inline void async(T &&lambda) {
     MARK_OVH(current_ws()->id);
     typedef typename std::remove_reference<T>::type U;
     hclib_async(lambda_wrapper<U>, new U(lambda), nullptr, nullptr, nullptr, 0);
 }
 
+/*
```
agrippa: I'd vote for deleting this comment entirely, since the rationalization is talking about HabaneroUPC++ which this code removes.

inc/hclib-place.h
```diff
@@ -48,10 +44,6 @@ typedef struct place_t {
 	int level; /* Level in the HPT tree. Logical root is level 0. */
 	int nChildren;
 	short type;
-#ifdef HC_CUDA
```
agrippa, line 51: Delete NVGPU_PLACE, AMGPU_PLACE, FPGA_PLACE, PGAS_PLACE in place_type_t above too?

inc/hclib-place.h
```diff
@@ -87,32 +79,9 @@ extern void *hclib_allocate_at(place_t *pl, size_t nbytes, int flags);
...
-        void *src, size_t nbytes, hclib_future_t **future_list,
-        void *user_arg, hclib_promise_t *out_promise);
-extern hclib_promise_t *hclib_async_memset(place_t *pl, void *ptr, int val,
-        size_t nbytes, hclib_future_t **future_list, void *user_arg);
-extern void hclib_async_memset_helper(place_t *pl, void *ptr, int val,
-        size_t nbytes, hclib_future_t **future_list, void *user_arg,
-        hclib_promise_t *out_promise);
-#endif
-
 inline short is_cpu_place(place_t * pl) {
```
agrippa, line 82: Is this used anywhere today? Seems like with these changes it would always return true, so maybe it can be deleted too?

src/hclib-hpt.c
```diff
@@ -287,181 +244,6 @@ inline void init_hc_deque_t(hc_deque_t *hcdeq, place_t *pl) {
...
-    hclib_promise_t *promise = hclib_promise_create();
-    hclib_async_memset_helper(pl, ptr, val, nbytes, future_list, user_arg,
-                              promise);
-    return promise;
-}
-#endif
-
 static const char *MEM_PLACE_STR   = "MEM_PLACE";
 static const char *CACHE_PLACE_STR = "CACHE_PLACE";
 static const char *NVGPU_PLACE_STR = "NVGPU_PLACE";
```
agrippa, line 249: Delete unused place strings here?

src/hclib-runtime.c
```diff
@@ -216,13 +168,8 @@ void hclib_entrypoint() {
      */
     hclib_global_init();
 
-#ifdef HC_COMM_WORKER
-    const int have_comm_worker = 1;
-#else
-    const int have_comm_worker = 0;
-#endif
     // init timer stats
-    hclib_initStats(hclib_context->nworkers, have_comm_worker);
```
agrippa, line 225: Looks like we can delete this second parameter too, and the static have_comm_worker in hclib-timer.c

agrippa (commented):
Just a few comments, otherwise looks great!

inc/hclib-async.hpp
DaoWen: I left the comment because it explains the rational weird name: `async_at_hpt`. I think we should leave it until we do something about the confusing name. The "at_hpt" suffix makes it sound like it takes an HPT argument rather than a place.

DaoWen: I'll clarify the comment (noting that the HCUPC++ function doesn't exist anymore) and mark it with a FIXME while I'm addressing the other comments).

agrippa: Sure, I think my concern is now that HUPC++ is ripped out of the code base this comment/rationale doesn't make any sense anymore because it lacks context. I would be okay with renaming this API as part of this commit (e.g. async_at_place), but I also don't really understand why we need comments talking about how much someone in the past subjectively disliked a particular API name. It's not great, but it's not bad enough that anyone has felt really motivated to change it in the last year.

inc/hclib-place.h
DaoWen: My goal with this patch was to remove the majority of the dead feature code that's cluttering up the codebase. While these types aren't used in the runtime anymore, they *are* used by the `hwloc_to_hpt` tool. I don't want to start messing with `hpt.dtd` in this commit. If you want to follow up with a commit that fixes all of those loose ends, I'd be fine with that—but I think you want to dump the XML descriptions in favor of your JSON machine description format somewhere down the road anyway, so maybe we can just leave the HPT's XML description as-is until we retire it.

agrippa: Okay, that sounds reasonable to me.

agrippa:
This pull request looks good to me, @budimlic 's turn to review and then merge.

budimlic (approved):
Looks good. Thanks for doing this, Nick!

PR #55, opened 2017-05-18
Fix C includes for C++ headers

DaoWen:
C headers should be included from inside the `extern "C"` block.


agrippa:
LGTM. Was there a particular compiler that triggered some errors, leading to this fix?

DaoWen:
I added a function prototype to one of the headers when I was working on something else, and that resulted in a warning about linkage mismatch (C vs C++).

PR #58, opened 2017-05-25
Restore fixed worker pool strategy

DaoWen:
This patch re-enables the option of using the original worker strategy with a fixed pool of thread contexts (no fibers). This option can be useful for debugging, performance profiling, and as a way to avoid overheads introduced by fibers when an application uses a compatible subset of the API.

Note that the different strategies now use C conditionals rather than preprocessor conditionals to select the strategy. This allows us to support either run-time selection (for debugging) or compile-time selection (for production) of different thread strategies. I think removing most of the preprocessor `#if`s has also made the code more readable and maintainable. When the strategy is selected statically, the compiler should remove all of the dead branches for the unused strategies, with a result very similar to the version using preprocessor conditionals.

A simple python script is included to make it easier to select a desired set of worker context options when building the runtime (if set statically for production mode) or when launching a program (if set dynamically for debug mode). There are a few placeholders for the `HCLIB_WORKER_STRATEGY_THREADS` strategy, which will be added in a follow-on patch. Another follow-on patch will update our Travis configuration to run our regression tests against a matrix of worker-context configurations.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,7 +12,7 @@ AUX_DIST = \
 library_includedir=$(includedir)
 
 library_include_HEADERS = inc/hclib-timer.h inc/hclib-rt.h inc/hclib.h \
-						  inc/hclib_config.h \
+						  inc/hclib_config.h inc/hclib-worker-config.h \
 						  inc/hclib-mak/hclib.mak inc/hclib-async.hpp src/inc/hclib-hpt.h \
 						  inc/hclib-forasync.hpp inc/hclib-promise.h inc/hclib-promise.hpp \
 						  src/inc/hclib-atomics.h inc/hclib-place.h \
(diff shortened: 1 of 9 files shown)


vivkumar:
Good to see that this option is not ripped out. I think this will be very useful while taking performance numbers where we want to limit HClib overheads to the minimal.

agrippa (commented):
Before I give this a detailed review, do we have some applications where not using fibers has measurable benefit? The fact is that several larger projects use fibers of some kind or another for flexibility (e.g. Realm), and so it would seem that for them fibers haven't been such a performance issue that they thought something like this was necessary. This is going to significantly increase the complexity of the code base, reduce its long-term maintainability, and just generally make it tougher to reason about. If the claim is that we're getting some improvement in performance for something we care about, my opinion is that the benefit needs to be demonstrated before this gets merged. I'm also unclear on how this helps debugging?

agrippa:
If you can tell me how to build this branch with the config you expect to perform the best, i can compare it against master with the micro-benchmarking suite.

PR #64, opened 2018-06-12
Add vector of dependencies to tasks.

srirajpaul:
Supports a maximum of 4 futures as dependency. To change maximum limit adjust the MAX_NUM_WAITS macro inside hclib-promise.h and rebuild HClib.

### DIFF
--- a/inc/hclib-async.h
+++ b/inc/hclib-async.h
@@ -35,6 +35,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *      Acknowledgments: https://wiki.rice.edu/confluence/display/HABANERO/People
  */
 #include <functional>
+#include <vector>
 
 #include "hclib.h"
 #include "hclib-async-struct.h"
@@ -148,6 +149,16 @@ inline hclib_task_t* _allocate_async(T *lambda) {
 }
 
 template <typename T>
+inline void async_await_at_helper(T&& lambda, hclib_future_t **futures,
+        const int nfutures, hclib_locale_t *locale, const int non_blocking) {
+    MARK_OVH(current_ws()->id);
+    typedef typename std::remove_reference<T>::type U;
+    hclib_task_t* task = initialize_task(call_lambda<U>, new U(lambda));
+    task->non_blocking = non_blocking;
+    spawn_await_at(task, futures, nfutures, locale);
+}
+
+template <typename T>
 inline void async(T &&lambda) {
 	MARK_OVH(current_ws()->id);
     typedef typename std::remove_reference<T>::type U;
@@ -189,6 +200,16 @@ inline void async_nb_await(T&& lambda, hclib_future_t *future) {
 }
 
 template <typename T>
+inline void async_nb_await(T&& lambda, std::vector<hclib_future_t *> &futures) {
+    async_await_at_helper(lambda, futures.data(), futures.size(), nullptr, 1);
+}
+
+template <typename T>
+inline void async_nb_await(T&& lambda, std::vector<hclib_future_t *> &&futures) {
+    async_await_at_helper(lambda, futures.data(), futures.size(), nullptr, 1);
+}
+
+template <typename T>
 inline void async_nb_await_at(T&& lambda, hclib_future_t *fut,
         hclib_locale_t *locale) {
     MARK_OVH(current_ws()->id);
@@ -196,6 +217,18 @@ inline void async_nb_await_at(T&& lambda, hclib_future_t *fut,
     hclib_task_t *task = initialize_task(call_lambda<U>, new U(lambda));
     task->non_blocking = 1;
     spawn_await_at(task, fut ? &fut : NULL, fut ? 1 : 0, locale);
+}
+
+template <typename T>
+inline void async_nb_await_at(T&& lambda, std::vector<hclib_future_t *> &futures,
+        hclib_locale_t *locale) {
+    async_await_at_helper(lambda, futures.data(), futures.size(), locale, 1);
+}
+
+template <typename T>
+inline void async_nb_await_at(T&& lambda, std::vector<hclib_future_t *> &&futures,
+        hclib_locale_t *locale) {
+    async_await_at_helper(lambda, futures.data(), futures.size(), locale, 1);
 }
 
 template <typename T>
@@ -244,6 +277,16 @@ inline void async_await(T&& lambda, hclib_future_t *future1,
 }
 
 template <typename T>
+inline void async_await(T&& lambda, std::vector<hclib_future_t *> &futures) {
+    async_await_at_helper(lambda, futures.data(), futures.size(), nullptr, 0);
+}
+
+template <typename T>
+inline void async_await(T&& lambda, std::vector<hclib_future_t *> &&futures) {
+    async_await_at_helper(lambda, futures.data(), futures.size(), nullptr, 0);
+}
+
+template <typename T>
 inline void async_await_at(T&& lambda, hclib_future_t *future,
         hclib_locale_t *locale) {
 	MARK_OVH(current_ws()->id);
@@ -272,6 +315,18 @@ inline void async_await_at(T&& lambda, hclib_future_t *future1,
 	spawn_await_at(task, futures, nfutures, locale);
 }
 
+template <typename T>
+inline void async_await_at(T&& lambda, std::vector<hclib_future_t *> &futures,
+        hclib_locale_t *locale) {
+    async_await_at_helper(lambda, futures.data(), futures.size(), locale, 0);
+}
+
+template <typename T>
+inline void async_await_at(T&& lambda, std::vector<hclib_future_t *> &&futures,
+        hclib_locale_t *locale) {
+    async_await_at_helper(lambda, futures.data(), futures.size(), locale, 0);
+}
+
 /*
  * Some CUDA compilers trip over the following line:
  *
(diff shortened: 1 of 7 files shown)


inc/hclib-async.h
```diff
@@ -188,6 +189,24 @@ inline void async_nb_await(T&& lambda, hclib_future_t *future) {
 	spawn_await(task, future ? &future : NULL, future ? 1 : 0);
 }
 
+template <typename T>
+inline void async_nb_await(T&& lambda, std::vector<hclib_future_t *> &futures) {
+	MARK_OVH(current_ws()->id);
```
agrippa: Looks like there's an alignment issue here. All lines should be indented by 4 spaces.

agrippa: Actually, looks like there's an alignment issue throughout this file. Since this is pre-existing, could you just file an issue saying that alignments are off in hclib-async.h to make sure I fix it?

inc/hclib-async.h
```diff
@@ -198,6 +217,26 @@ inline void async_nb_await_at(T&& lambda, hclib_future_t *fut,
     spawn_await_at(task, fut ? &fut : NULL, fut ? 1 : 0, locale);
 }
 
+template <typename T>
+inline void async_nb_await_at(T&& lambda, std::vector<hclib_future_t *> &futures,
```
agrippa, line 223: Is there a reason futures has type std::vector<hclib_future_t *>& here and std::vector<hclib_future_t *>&& in other places?

agrippa (changes requested):
Sriraj,

I appreciate the usefulness of vector-based routines, but I think we need to cut down on the amount of copy-pasting in this code change before I finish my review. Otherwise it'll be a bit difficult to maintain in the long term. It seems pretty feasible that many of these vector-based routines could simply call their non-vector-based cousins in this same file, rather than copy-pasting them. Would you agree?

I'm also a bit unclear on how some of these variants differ. Could you explain when I would want to use:

async_future_await_at(T&& lambda, std::vector<hclib_future_t *> &&futures, hclib_locale_t *locale)

vs.

async_future_await_at(T&& lambda, std::vector<hclib_future_t *> &futures, hclib_locale_t *locale)

srirajpaul:
Max, 
I am not quiet sure how the vector based functions can be implemented by calling the non-vector-based cousins in this same file. Say for example, if we compile with MAX_NUM_WAITS=7 since we had 7 dependencies for our 3D stencil, there are no non-vector based function that accepts 7 futures. Or we need to create Variadic templates, and iterate over them and create an array and so on to support it. Current implementation seems to be much more readable/concise than that. Also getting array pointer directly from vector is efficient/constant time whereas creating array for vardic function is linear and not so efficient.

A) std::vector<hclib_future_t *> &futures is lvalue reference and 
B) std::vector<hclib_future_t *> &&futures is rvalue reference. 
An example would be,
vector<hclib_future_t *> fut_vec{f1, f2, f3};  async_await(...,fut_vec); will invoke A whereas,
async_await(..., vector<hclib_future_t *>{f1, f2, f3}); would invoke B

Prior to c++11, there was no rvalue reference and therefore such invocations used call by value and therefore was expensive due to the copy.

There is a way to combine them which is the universal reference which can take both  lvalue reference and  rvalue reference  (similar to T&& lambda). But we need to convert 'vector<hclib_future_t *> &futures'  to something like  'V &&futures', where V is a template parameter and may be use std::enable_if to tell that V is of type std::vector<hclib_future_t *>.
Do you think converting to universal reference is a better option?

inc/hclib-async.h
srirajpaul: & is lvalue reference and && is rvalue reference.

inc/hclib-async.h
srirajpaul: I will add an issue. It seems that this file used tab instead of space and based on local tab configuration it will show different spacing.

inc/hclib-async.h
agrippa: Sure, I understand that, but why have both? What's the reason?

agrippa:
Sriraj, so why not just add some non-vector-based routines that both the vector- and non-vector-based routines call (thereby removing the copy-paste effect)?

For example, you've duplicated async_future_await a couple of times. Why not just add:

 async_future_await(T&& lambda, hclib_future_t** futures)

and have both the existing and new routines call this directly, basically turning both new and existing into one-liners sharing a common implementation? It seems like you should be able to do this for all of the added code.

Let's first make that change, then see what that collapses the rvalue/lvalue versions down to and see if the code sprawl is sufficiently mitigated to leave them both in place.

inc/hclib-async.h
srirajpaul: vector<hclib_future_t *> fut_vec{f1, f2, f3}; async_await(...,fut_vec); will invoke lvalue reference whereas,

async_await(..., vector<hclib_future_t *>{f1, f2, f3}); would invoke rvalue reference. If function with &&futures is not defined,  it will give compilation error since there is no function matching this signature.

srirajpaul:
Added the helper function that takes an array of dependencies and the new routines go through it. Old routines  can also use this helper function, but prefer not to make changes to existing routines in this pull request.

inc/hclib-promise.h
```diff
@@ -59,7 +59,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 /*
  * The maximum number of future objects a task can wait on.
  */
+#ifndef MAX_NUM_WAITS
 #define MAX_NUM_WAITS 4
```
agrippa: Could you explain how you see this being used? Is this being added as a configure-time option?

agrippa:
These changes look good! Just one additional question I added.

inc/hclib-promise.h
srirajpaul: We need to build HClib with '-DMAX_NUM_WAITS=N' CXXFLAG.
For example to set the number to 6, build HClib as,
'HC_FLAGS_1=-DMAX_NUM_WAITS=6 ./install.sh'

Added as a part of the comment for this checkin.

agrippa: Can we just add this as a configure flag then? I didn't even realize that HC_FLAGS crud was still in there, it should be ripped out and we certainly shouldn't be adding more stuff that relies on it.

srirajpaul: Added a configure flag --with-max-dependency=n. It will show up with a brief explanation when using configure --help.

agrippa: Great! Can we delete the define now? For example, something like:

#ifndef MAX_NUM_WAITS
#error MAX_NUM_WAITS is not defined
#endif

srirajpaul: Why do we need to do that?
Currently we are using this default value set using #define if the user do not provide the  --with-max-dependency while building hclib. 
Also hclib.h includes hclib-promise.h, in which MAX_NUM_WAITS is defined and used, and therefore a user who uses hclib.h will get this error message if they dont define  MAX_NUM_WAITS while compiling their application.  
Otherwise we need to move code around so that those parts of the code not used by user should be invisible from the user.

agrippa: I agree that we need MAX_NUM_WAITS defined at all times. I didn't realize --with-max-dependency didn't provide a default value if it wasn't set. Is it possible to do that?

My concern is if HClib is built with one value of MAX_NUM_WAITS, and then an application is compiled with a different value #defined for MAX_NUM_WAITS, that conflict will be difficult to debug. Optimally, we want a single value set for MAX_NUM_WAITS at configure time which can't overridden accidentally at application compile time (or at least, if it is it requires someone actually going in and deleting code in header files, in which case we have to assume they know what they're doing).

Makes sense?

srirajpaul: That makes sense.  --with-max-dependency can provide default value but those values are available only during building HClib and will be used inside libhclib. Later when hclib.h is included in application program, it will  not get those macro values given during  building HClib (those values are only inside libhclib).
But I dont see any easy solution around that other than performing refactoring so that only values/structs etc required by the application programmer comes in exposed header files and others gets moved to some internal headers/source. For now can we change 

#ifndef MAX_NUM_WAITS
#define MAX_NUM_WAITS 4
#endif

to just 
#define MAX_NUM_WAITS 4

This way there is only point where MAX_NUM_WAITS is defined and for now if anyone wants to change it they need to edit the file and recompile, and create a bug/feature request to refactor the code?

agrippa: Wait, wouldn't that change just make --with-max-dependency unnecessary again?

srirajpaul: Yes. I will remove it, other wise it will give compilation error that MAX_NUM_WAITS is redefined.

agrippa: Okay, so then the plan for any projects that need this > 4 (which I imagine yours might?) would be for them to edit this manually?

I'm fine with that, just want to make sure I'm understanding correctly.

The only big pain point on that front would be whenever we want to merge code in again. Any edits to MAX_NUM_WAITS might (or might not) have to be reverted first to keep it consistent on the master branch.

One future exploration might be to look at how this value affects microbenchmarks.

srirajpaul: Yes. If someone needs to increase, then they have to edit it manually and we need to explore how it affect the benchmarks when that value is increased.
Also In the resilience branch I am making the number of futures a field within the task struct so that there is no need to set the hardcoded value and array size can be variable. This is because, for sparse matrix problems the number of dependencies are not uniform and they can vary from 0 to very large numbers. We do not want to allocate such large dependency array for all tasks when it is not needed. But we need to find out how much such a change affects performance before making any such changes in this branch.

agrippa: Okay, sounds good. Can you make the change to delete these #ifndefs and the added configure stuff?

Once that's changed this looks good to me, but Zoran will have to sign off too (every merge requires 2 pairs of eyeballs). He also still needs to look at Akihiro's pull request.

@budimlic back from vacation yet?

srirajpaul: Great. Removed the ifndefs and configure flags.

budimlic:
I'm OK with this pull request as a quick fix to allow tasks to wait on more than 4 futures . 

However, it seems like we should now (with the APIs added by Sriraj allowing a vector of futures as an argument of async_await) be able to get rid of the MAX_NUM_WAITS completely. It's only used in hclib_task_t struct  to allocate a NULL-terminated array of futures on which the task is waiting on. If this is done for performance reasons (I suspect it is), we could keep doing it for a fixed number of futures, then store any additional future pointers in a dynamically allocated array if needed. This would require storing the number of futures on which the task awaits into the hclib_task_t struct, which I would prefer to the NULL-terminated array kludge we have now anyway.  

Requiring a user to recompile hclib to allow for more than 4 futures seems very counter-intuitive now that we have a vector-of-futures API.

agrippa:
@budimlic This change doesn't actually change anything in regards to how many futures we can wait on (that's a compile-time constant before this change, and after). It just adds API wrappers that use std::vector.

Using a dynamically allocated array extension is certainly something someone could submit as the next pull request.

agrippa:
@budimlic If you're ready, we've both signed off now and this can be merged in.

agrippa (approved):

budimlic (approved):

budimlic:
@agrippa Unless I'm missing something obvious, there was no way until now to create an async_await that depends on 5 futures, even if I compiled with MAX_NUM_WAITS=5

agrippa:
@budimlic You would just use the C APIs.

budimlic:
Good point. So this change brings up the expressiveness of the C++11 API  for async_await up to the level of expressiveness of the C API, at the cost of exposing the MAX_NUM_WAITS to the C++11 programmer. Since MAX_NUM_WAITS is already exposed to the C programmer, there is no downside to this change.

Getting rid of MAX_NUM_WAITS would be a topic of separate pull request.

PR #66, opened 2018-12-15
Restructure installation process to place everything under a single install directory

agrippa:
This change includes edits to the README to clarify the installation process, changes to installation process for modules so that they just go under the same HCLIB_ROOT directory, and updates to the code so that it looks for modules under $HCLIB_ROOT.


srirajpaul:
Changes look good. 

Few suggestions. 
Inside install.sh, the new added modules/system installation currently fails in a fresh system since HCLIB_ROOT is not set. This is happening with travis now.
may be something like 'HCLIB_ROOT=${INSTALL_PREFIX} make install' would work.

Should we add the modules/system building in README.md ? Currently no Hclib program can without it.

Also we can remove the '- pushd modules/system && make && popd' from .travis.yml since it is taken care inside install.sh

agrippa:
@srirajpaul good catch on the fresh install problem, I've fixed this in install.sh.

I've also added a short section on the module installation process in README.md.

srirajpaul:
Looks good !!!

PR #67, opened 2018-12-18
adding basic examples for upcoming HiPC tutorial

vivkumar:


srirajpaul:
Overall looks good. Few comments.

In case we are are planning for more extensive release of HClib, can we have a more hierarchical directory structure such as /test/tutorial/hipc18/cpp/... Or we can also keep the current structure if we plan to use almost the same tutorial. 

Also as we discussed we can use wait() instead of wait_and_get() to make it future proof when wait_and_get gets removed. 

test/tutorial/cpp/futures/graph.cpp many get calls, for example Line 28 a->get() does not wait for Task A to finish since futures do nothing w.r.t get(). Might want to use wait() there if we want to create a graph.

test/tutorial/cpp/asyncfinish/Makefile
```diff
@@ -0,0 +1,17 @@
+include $(HCLIB_ROOT)/include/hclib.mak
```
agrippa: Lots of questions on this makefile (and the others that mirror it):

1. Why does clean-ob/clean remove .dSYM? We shouldn't be creating any libraries, right? Or is this some mac-specific file I'm not familiar with?
2. Same question, but for *.o. Doesn't seem like we're creating object files.
3. Why do we need both clean and clean-obj? Seems like just having clean is good enough.
4. Why does 'all' trigger 'clean' and 'clean-obj'? I'd personally prefer just 'all: $(EXE)'

test/tutorial/cpp/asyncfinish/arraysum.cpp
```diff
@@ -0,0 +1,55 @@
+#include "hclib_cpp.h"
+#include <sys/time.h>
+
+int SIZE=10000000; //10M
+int* array;
+
+// TIMING HELPER FUNCTIONS
+long get_usecs ()
```
agrippa: We already have hclib_current_time_ns() and hclib_current_time_ms(). Could you replace all timer utilities in the application code to use one of these HClib functions?

test/tutorial/cpp/ddt/graph.cpp
```diff
@@ -0,0 +1,39 @@
+#include "hclib_cpp.h"
+#include <unistd.h>
+
+/*
```
agrippa: I think if this material is intended to be used in tutorials, every file should have a descriptive comment like the one below at the top.

Not necessarily with a picture of the task graph, but at a minimum there should be a comment describing what the file implements (even for files that have a semi-descriptive filename, e.g. "fib").

test/tutorial/cpp/forasync/averaging.cpp
```diff
@@ -0,0 +1,102 @@
...
+        myNew[j] = (initialOutput[j - 1] + initialOutput[j + 1]) / 2.0;
+    }
+    double* temp = myNew;
+    myNew = initialOutput;
+    initialOutput = temp;
+  }
+}
+
+void runParallel() {
+  // note we are not using ceil(...) as we did in async-finish version.
```
agrippa: Can we use /* */ style comments instead of // ?

agrippa (changes requested):
Comments are inlined. I would also support Sriraj's comment about the folder structure, and would suggest we just create a top-level tutorials/ folder with this code underneath it in a folder called hipc18

agrippa:
@vivkumar what's the status of this? If I remember correctly, the tutorial is on the 20th, which means we're cutting this very tight. It's important to keep in mind the time zone that Sriraj and I are working on, in getting reviews back to you.

vivkumar:
Apologies for delay. I am at HiPC and did not get time to look into this earlier. I just committed the updates.

agrippa:
@vivkumar some final comments, mostly this looks great!

1. Can we just move this entirely out of the test/ folder? These aren't really tests per se, so I'd rather just have a top-level tutorials/ folder (at the same level as test/) with a single hipc18 subfolder.
2. I'm not sure why but one of the tests failed (doesn't seem like this is because of something in the code?). Hopefully, all the tests will pass with the directory structure change.

tutorial/hipc18/cpp/futures/graph.cpp
```diff
@@ -0,0 +1,59 @@
...
+int main(int argc, char **argv) {
+  char const *deps[] = { "system" }; 
+  hclib::launch(deps, 1, [&]() {
+    hclib::future_t<void> *a = hclib::async_future([=]() {
+      sleep(1);
+      printf("A\n");
+      return;
+    });
+    hclib::future_t<void> *b = hclib::async_future([=]() {
+      a->get();
```
srirajpaul: This get() will not wait for Task A to finish. We should use wait() here. Similarly all the get() calls in this example.

srirajpaul (changes requested):

srirajpaul (approved):

agrippa (approved):

PR #69, opened 2018-12-19
Fix compilation issue on OSX and remove modules/system folders from build

srirajpaul:
-force_load option is used to load static library in OSX and it was used with JSMN. Since JSMN is not a static library anymore, no need to load it.

modules/system libraries and include files are copied to HCLIB_ROOT and therefore no need to add these extra paths in building HClib programs.

### DIFF
--- a/inc/hclib-mak/hclib.mak
+++ b/inc/hclib-mak/hclib.mak
@@ -7,17 +7,13 @@ HCLIB_CFLAGS=-I$(HCLIB_ROOT)/include
 HCLIB_CXXFLAGS=-std=c++11 $(HCLIB_CFLAGS)
 HCLIB_LDFLAGS=-L$(HCLIB_ROOT)/lib
 
-ifndef GET_LINK_FLAG
-	GET_LINK_FLAG = -Wl,$(1)
-endif
-
 UNAME_S := $(shell uname -s)
 ifneq ($(UNAME_S),Darwin)
     IS_MAC_OS = 0
     HCLIB_LDLIBS=-lhclib -lrt -ldl
 else
     IS_MAC_OS = 1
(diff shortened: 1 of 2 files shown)


inc/hclib-mak/hclib.mak
```diff
@@ -7,17 +7,17 @@ HCLIB_CFLAGS=-I$(HCLIB_ROOT)/include
 HCLIB_CXXFLAGS=-std=c++11 $(HCLIB_CFLAGS)
 HCLIB_LDFLAGS=-L$(HCLIB_ROOT)/lib
 
-ifndef GET_LINK_FLAG
-	GET_LINK_FLAG = -Wl,$(1)
-endif
+#ifndef GET_LINK_FLAG
```
agrippa: Can we just delete these lines, rather than commenting them?

modules/system/inc/hclib_system.post.mak
```diff
@@ -1,8 +1,8 @@
 # Post Makefile includes are the main part of a module's build system, allowing
 # it to add flags to the overall project compile and link flags.
-HCLIB_CFLAGS+=-I$(HCLIB_ROOT)/../modules/system/inc
-HCLIB_CXXFLAGS+=-I$(HCLIB_ROOT)/../modules/system/inc
-HCLIB_LDFLAGS+=-L$(HCLIB_ROOT)/../modules/system/lib
+#HCLIB_CFLAGS+=-I$(HCLIB_ROOT)/../modules/system/inc
```
agrippa: Can we just delete these lines, rather than commenting them?

agrippa (changes requested):
Just some tiny changes.

srirajpaul:
Removed the unused lines rather than commenting them out.

PR #72, opened 2019-02-13
Add text to README explicitly requiring c11

agrippa:


agrippa:
@srirajpaul could you just check this out quickly?

srirajpaul:
Looks good.

PR #74, opened 2019-06-06
Add support for Honeycomb simulator

### LINKED ISSUE #71: Add support for Honeycomb simulator
Adding support for Honeycomb simulator requires adding the assembly support for context switching and conditional compilation for a few unsupported features.

srirajpaul:
resolves #71 

Add the context switching assembly for honeycomb.
Make unsupported libraries (e.g. dlopen) compile only conditionally.
Add honeycomb architecture to Configure/Makefile scripts.


srirajpaul:
@agrippa Please take a look.

src/hclib-instrument.c
```diff
@@ -4,13 +4,30 @@
...
 #include <signal.h>
 #include <errno.h>
-#include <aio.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 
 #include "hclib.h"
 #include "hclib-instrument.h"
 
+#ifdef HAVE_AIO_H
```
agrippa, line 13: @srirajpaul I'm not a autoconf expert. Am I correct in assuming that HAVE_AIO_H will be set so long as autoconf was able to find aio.h during configuration? So on most systems, it would be set?

src/hclib-runtime.c
```diff
@@ -56,6 +55,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <hclib-module.h>
 #include <hclib-instrument.h>
 
+#ifdef HAVE_DLFCN_H
```
agrippa, line 58: @srirajpaul where is HAVE_DLFCN_H set?

src/inc/litectx.h
```diff
@@ -14,7 +14,10 @@
 #include "hclib_common.h"
 #include "fcontext.h"
 #include <string.h>
+
+#ifdef HAVE_SYS_MMAN_H
```
agrippa, line 18: I'm assuming we have the same situation here as with aio.h? HAVE_SYS_MMAN_H will be automatically set if sys/mman.h is findable on a given system, by autoconf?

src/hclib-instrument.c
srirajpaul: @agrippa  That is correct. HAVE_AIO_H gets set as  "#define HAVE_AIO_H 1" in hclib-install/include/hclib_config.h if autoconf is able find aio.h in the target platform. In most systems it will be set.

src/hclib-runtime.c
srirajpaul: @agrippa  set inside  hclib-install/include/hclib_config.h when HClib is build.

src/inc/litectx.h
srirajpaul: @agrippa  Yes. it is set inside  hclib-install/include/hclib_config.h when HClib is build.

src/hclib-runtime.c
agrippa: @srirajpaul is this a built-in autoconf thing? I'm only asking because I see you've added special commands for AIO but not for this. I realize this is probably a dumb question due to my autoconf ignorance, just want to make sure we aren't missing anything

srirajpaul: @agrippa  DLFCN was already supported, as in the installed hclib-install/include/hclib_config.h was already setting HAVE_DLFCN_H to 1 if dlfcn.h is present.

This was happening because https://github.com/habanero-rice/hclib/blob/32ada897f673edd56375e83a58c1a8cfd5ba5dcd/configure.ac#L21 was using a macro AC_PROG_LIBTOOL to enable support for shared libraries. This was automatically adding the AC_CHECK_HEADERS([dlfcn.h]..) and together with HAVE_DLFCN_H  in https://github.com/habanero-rice/hclib/blob/32ada897f673edd56375e83a58c1a8cfd5ba5dcd/inc/hclib_config.h.in#L7 sets the HAVE_DLFCN_H macro to 1 in hclib_config.h  if dlfcn.h is present.
The question is not dump. I am also learning more about autoconf through this and it took some time to find out how DLFCN was getting set.

Issue #73, opened 2019-06-03
Deadlock when using finish and async_await


srirajpaul:
When using a sequence of async_await finish async_await , they dead lock intermittently.

A sample code that produces the error is as follows
------
```
  #include "hclib_cpp.h"

  int main (int argc, char *argv[])
  {
      int n_rounds = 8, n_tiles=5;
      const char *deps[] = { "system" };
      hclib::launch(deps, 1, [&]() {

          //create a 2D array of promises
          auto tile_array = new hclib::promise_t<void*>**[n_rounds+1];
          for (int tt=0; tt < n_rounds+1; ++tt) {
              tile_array[tt] = new hclib::promise_t<void*>*[n_tiles];
              for (int j=0; j < n_tiles; ++j) {
                  tile_array[tt][j] = new hclib::promise_t<void*>();
              }
          }

            for (int j=0; j < n_tiles; ++j)
                tile_array[0][j]->put(nullptr);

            for (int tt=0; tt < n_rounds; ++tt){
              for (int j=0; j<n_tiles; ++j){
                int left_nbr = (n_tiles+j-1)%n_tiles;
                int right_nbr = (j+1)%n_tiles;
                hclib::async_await([=]{
                  hclib::finish([=]{
                    hclib::async_await([=]{
                        tile_array[tt+1][j]->put(nullptr);
                    }, tile_array[tt][left_nbr]->get_future());

                  });

                }, tile_array[tt][left_nbr]->get_future(),
                   tile_array[tt][right_nbr]->get_future());
              }
            }

      });
  }
```
-----
to run the program use the commands:
```
export HCLIB_WORKERS=2
for i in {1..1000}
do
    echo $i
    ./a.out
done
```

agrippa:
Confirmed this is reproducible with provided example, simply running repeatedly will lead to deadlock.

Attaching gdb doesn't reveal any particularly interesting stacks. Mostly just:

#0  0x00007fcce8feafc8 in locale_steal_task (ws=ws@entry=0x13d1310, stolen=stolen@entry=0x7fccd4040a00, out_victim=out_victim@entry=0x7fccd40409fc)
    at ../../src/hclib-locality-graph.c:877
#1  0x00007fcce8fe4a89 in find_and_run_task (on_fresh_ctx=1, flag_val=0, current_finish=0x0, flag=<optimized out>, ws=0x13d1310) at ../../src/hclib-runtime.c:626
[...]

PR #77, opened 2019-06-11
Fix issue #73: Deadlock when using finish and async_await

sbak5:
This fixes the issue #73 considering that root finish can run in core_work_loop so all the workers cannot  get out of the loop by 'hclib_signal_join'

### DIFF
--- a/inc/hclib-async-struct.h
+++ b/inc/hclib-async-struct.h
@@ -47,6 +47,7 @@ extern "C" {
 #endif
 
 extern void spawn(hclib_task_t * task);
+extern void spawn_root(hclib_task_t * task);
 extern void spawn_await_at(hclib_task_t *task, hclib_future_t **futures,
         const int nfutures, hclib_locale_t *locale);
 extern void spawn_at(hclib_task_t *task, hclib_locale_t *locale);
--- a/inc/hclib-rt.h
+++ b/inc/hclib-rt.h
@@ -108,6 +108,11 @@ typedef struct _hclib_worker_state {
      * Information on currently executing task.
      */
     void *curr_task;
+    /* Indicate whether this work is initiating hclib_finalize */ 
+    int is_terminator;
+
+    /* This value is used to get ouf of the core_work_loop and find_and_run_task under the root finish */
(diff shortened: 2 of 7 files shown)


agrippa:
@sbak5 thanks for taking a look at this -- I'm sure this has been a tough one to think through.

Before opening pull requests, it's always a good idea to run all of the c and cpp tests we have to verify no regressions have been introduced. I just tried that, and it looks like there's now a failing assertion in cpp/finish0. There may be other failing tests, I only took a quick look. I believe the c/memory/allocate test is now hanging as well (see the Travis builds, which hit that error). We'll need those issues resolved before merging (as well as any other failing tests), as both of those tests pass on the master branch.

You should be able to simply use the test_all.sh scripts under the test/c/ and test/cpp/ directories to run all of the tests.

Thanks!

Max

sbak5:
Hello, Max,
Yeah, my patch doesn't consider the example where future is used. 
I'll push a patch to resolve the case to resolve the issue. 

Thanks,
Seonmyeong Bak

sbak5:
I've pushed a fix to resolve the build errors. 

The main issue was that the termination of runtime was not synchronized. 

I added a task which is dependent on the root task so that the task signals other workers the root task is completed (now the counter value of the root finish is final). After this signal, we can use the counter value of the root finish as our termination condition.

agrippa:
@sbak5 looks good, all the tests are passing on my laptop as well.

one final step: could you add Sriraj's original deadlock example as a test in the appropriate directory (I can't remember if it was c or cpp)? That way, we prevent against future regressions.

Thanks for working on this!

sbak5:
I just added. The name of the test is nested_finish_async_await.cpp in test/cpp

agrippa:
@sbak5 looks like the build failed/hung? could you try rerunning the Travis tests?

https://travis-ci.org/habanero-rice/hclib/jobs/545539336

sbak5:
How can I rerun the travis test? Do I need to recommit?

sbak5:
@agrippa The build of this commit is OK this time.

PR #78, opened 2019-06-19
Roll back the previous patch to fix the deadlock with refined fix

sbak5:
This request rolls back the previous commits to fix the deadlock (issue #73 ) and contains a fix to solve the issue completely.


agrippa:
This looks good to me. Just a note, as part of my local testing I actually hit this bug in an existing test as well (promise/future2).

PR #79, opened 2020-02-23
Expand CUDA module APIs, fix locality graph out-of-bounds

agrippa:
Some small extensions to the CUDA APIs to support waiting on multiple futures.

### DIFF
--- a/locality_graphs/summit.json
+++ b/locality_graphs/summit.json
@@ -1,0 +1,16 @@
+{
+    "nworkers": 1,
+    "declarations": [
+        "sysmem",
+        "GPU0"
+    ],
+    "reachability": [
+        ["sysmem", "GPU0"]
+    ],
+    "pop_paths": {
+        "default": ["sysmem", "GPU0"]
+    },
+    "steal_paths": {
+        "default": ["sysmem", "GPU0"]
+    }
+}
--- a/modules/cuda/inc/hclib_cuda.h
+++ b/modules/cuda/inc/hclib_cuda.h
@@ -18,27 +18,30 @@ __global__ void simd_driver_kernel(functor_type functor) {
     functor(tid, width);
 }
 
-template<class functor_type, class await_type>
+template<class functor_type>
 inline hclib::future_t<void> *forasync_cuda(const int blocks_per_gridx,
         const int blocks_per_gridy, const int blocks_per_gridz,
         const int threads_per_blockx, const int threads_per_blocky,
         const int threads_per_blockz, const int shared_mem,
         functor_type functor, hclib::locale_t *locale,
-        hclib::future_t<await_type> *future) {
+        std::vector<hclib_future_t *> &futures) {
     HASSERT(locale->type == get_gpu_locale_id());
 
     hclib::promise_t<void> *prom = new hclib::promise_t<void>();
 
     hclib::async_at([locale, functor, blocks_per_gridx, blocks_per_gridy,
             blocks_per_gridz, threads_per_blockx, threads_per_blocky,
-            threads_per_blockz, shared_mem, prom, future] {
+            threads_per_blockz, shared_mem, prom, futures] {
 #ifdef HCLIB_INSTRUMENT
     const unsigned _event_id = hclib_register_event(get_cuda_kernel_event_id(),
             START, -1);
 #endif
 
-        if (future) {
-            future->wait();
+        for (auto i = futures.begin(), e = futures.end(); i != e; i++) {
+            hclib_future_t *f = *i;
+            if (f) {
+                hclib_future_wait(f);
+            }
         }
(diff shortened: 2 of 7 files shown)


agrippa:
@srirajpaul could you take a look at this small PR before I merge?

srirajpaul:
@agrippa  Sure. I never used HClib with GPU. Will use this pull request as an excuse to start using HClib with CUDA.

locality_graphs/summit.json
```diff
@@ -0,0 +1,16 @@
+{
```
srirajpaul, line 1: Could name this as sample_gpu or some generic name with gpu rather than summit, so that it will be easy to find for someone new who want to try gpu.

modules/cuda/inc/hclib_cuda.h
```diff
@@ -18,27 +18,30 @@ __global__ void simd_driver_kernel(functor_type functor) {
...
 
-template<class functor_type, class await_type>
+template<class functor_type>
 inline hclib::future_t<void> *forasync_cuda(const int blocks_per_gridx,
         const int blocks_per_gridy, const int blocks_per_gridz,
         const int threads_per_blockx, const int threads_per_blocky,
         const int threads_per_blockz, const int shared_mem,
         functor_type functor, hclib::locale_t *locale,
-        hclib::future_t<await_type> *future) {
+        std::vector<hclib_future_t *> &futures) {
```
srirajpaul, line 27: Just for clarification. You are not adding the && version for now and will add when need arises.

src/hclib-fptr-list.c
```diff
@@ -40,7 +40,11 @@ void hclib_register_func(hclib_fptr_list_t **list, int index, void *fptr,
 
 void *hclib_get_func_for(hclib_fptr_list_t *list, int index) {
     assert(list);
-    return (list->fptrs)[index];
+    if (index < list->capacity) {
```
srirajpaul, line 43: is it expected we get an index greater than capacity, or will this push the bug from here to somewhere else?

src/hclib-locality-graph.c
```diff
@@ -1141,8 +1141,8 @@ hclib_locale_t *hclib_get_closest_locale_of_types(hclib_locale_t *locale,
     int to_visit_index = 0;
     int *to_visit = (int *)malloc(sizeof(int) * n_locales);
     hclib_locale_t *curr = locale;
-    while (!contains(curr->type, locale_types, n_locale_types) &&
-            visiting_index <= n_locales) {
+    while (visiting_index <= n_locales &&
```
srirajpaul: Look like the out of bounds transferred to another array. When data is not found, then visiting_index is n_locales+1. So inside the loop, in line 1155, this would access to_visit[n_locales] which might cause segmentation fault since to_visit is only of size n_locales. Either increase size of to_visit by 1 or better make the loop only access up to the legal range of 0 to n_locales-1

srirajpaul (changes requested):

locality_graphs/summit.json
agrippa: I'd rather not just have generic files in here -- if people don't know how to use them, then trying to use some generic gpu.json file would be more likely to lead to confusing behavior than anything else.

agrippa: Plus, I don't think anyone other than me is using these files anyway...

src/hclib-locality-graph.c
agrippa: Good point. I've just pushed another commit that should address this problem with this function.

src/hclib-fptr-list.c
agrippa: I think we should keep this behavior. Because these lists are lazily expanded, it is possible to get an index that is semantically valid but which the list hasn't been expanded to include yet (in which case the proper behavior would be to return the dummy value, NULL).

src/hclib-locality-graph.c
```diff
@@ -1151,12 +1158,9 @@ hclib_locale_t *hclib_get_closest_locale_of_types(hclib_locale_t *locale,
                 to_visit[to_visit_index++] = i;
```
srirajpaul, line 1158: Since to_visit_index starts from 1, to_visit[to_visit_index++] can go to out of bounds in case the the if condition is true on all iterations. If that case does not happen, then it is good to go.

srirajpaul (changes requested):

src/hclib-locality-graph.c
agrippa: This should never be possible, because there are at most n_locales and the contains(i, to_visit, to_visit_index) validates that this list does not already contain this item. So even though we may evaluate the conditional around this stement more than n_locales times, we will only ever hit this conditional n_locales-1 times.

srirajpaul: Good to merge.

srirajpaul (approved):

PR #80, opened 2020-02-26
Hclib mem tie break

agrippa:
@srirajpaul one more small bug fix.

When copying between two HClib places/locales, the runtime needs to figure out what kind of copy API it has to use (e.g. memcpy, cudaMemcpy, etc). If it encounters a copy between two locales that each say HClib must use its copy API, HClib has no way of really negotiating between the two and will abort. This change handles the trivial case where the two places being copied between both say HClib must use their proprietary API, but those two places are both asking for the same API.


srirajpaul (approved):
Looks good. Just remove the comment "Not both MUST_USE" or move it inside else. Go ahead and merge.

PR #83, opened 2020-05-12
Some new async APIs, small formatting and comments changes

agrippa:


agrippa:
@srirajpaul could you take a quick look at this and let me know if it looks good? most of this is formatting and comments, but there are a few new async APIs.

inc/hclib-async.h
```diff
@@ -403,15 +425,31 @@ auto async_future_await(T&& lambda, hclib_future_t *future) ->
...
 template <typename T>
 auto async_future_await(T&& lambda, std::vector<hclib_future_t *> &&futures) ->
         hclib::future_t<decltype(lambda())>* {
-    return async_future_await_at_helper(lambda, futures.data(), futures.size(), nullptr, 0);
+    return async_future_await_at_helper(lambda, futures.data(), futures.size(),
+            nullptr, 0);
 }
 
+template <typename T>
+auto async_nb_future_await(T&& lambda, std::vector<hclib_future_t *> &&futures) ->
```
srirajpaul, line 448: @agrippa I think this will not allow to pass a vector variable as shown below which would require '& futures' version. Just to confirm, is that intended?

vector< hclib_future_t*> v;
..
async_nb_future_await(.., v);

modules/common/hclib-module-common.h
```diff
@@ -39,11 +39,11 @@ void poll_on_pending(pending_op **addr_of_head,
                      * wait_set and which should be updated.
                      */
                     pending_op *old_head = __sync_val_compare_and_swap(
-                            addr_of_head, op, op->next);
+                            addr_of_head, op, next);
```
srirajpaul, line 42: @agrippa Just to confirm. Using op->next was also correct. Is it so?

srirajpaul (changes requested):

inc/hclib-async.h
agrippa: @srirajpaul i think that's a good point, i've added a reference variant of this API as well.

modules/common/hclib-module-common.h
agrippa: yep, it still worked. i'm just trying to be more consistent in this function so that the next item is always referred to through the next variable.

srirajpaul (approved):
Looks good. Please go ahead and merge.

PR #84, opened 2020-06-13
Add support for variable number of futures to be passed to async_await

agrippa:
@srirajpaul Could you take a look at this?


src/hclib-promise.c
```diff
@@ -179,6 +179,16 @@ int register_on_all_promise_dependencies(hclib_task_t *wrapper_task) {
         }
     }
 
+    if (wrapper_task->waiting_on_extra) {
+        while (wrapper_task->waiting_on_extra[wrapper_task->waiting_on_index - MAX_NUM_WAITS + 1]) {
+            wrapper_task->waiting_on_index++;
+            hclib_future_t *curr = wrapper_task->waiting_on_extra[wrapper_task->waiting_on_index - MAX_NUM_WAITS];
+            if (_register_if_promise_not_ready(wrapper_task, curr)) {
```
srirajpaul, line 187: should we add the  if (curr) check as above since nullptr is allowed to be passed as dependency?

srirajpaul (commented):
@agrippa Should we keep these additions inside a conditional compilation (#ifdef) since this might directly impact performance? It might not have big impact on actual applications since it won't incur much overhead if not used, but might show up in the task-microbenchmark-suite.

src/hclib-promise.c
agrippa: I'm not sure I would really consider that desirable behavior, and would actually break the code in this case as it is assuming that the list is NULL-terminated. This is nice as it means we don't have to add another field to the task structure and have it take up more space for tasks that don't use futures.

We could still add a field that is the # of futures. But a better question is, do you know if any of our benchmarks or tests currently assume they can pass a NULL pointer as part of the future list and it'll just be ignored? Do you disagree, and think that we should support this?

agrippa:
@srirajpaul on the #ifdef: I think that sounds reasonable. Question is, do we enable or disable arbitrary numbers of futures by default?

I'd argue to keep it enabled by default as it's the most flexible option for new users/tutorials. Then, if more advanced people want to turn it off to get better performance and smaller task structs they can figure that out. But curious to hear what you think.

src/hclib-promise.c
srirajpaul: In that case, let us keep it as a null-terminated list so that we do not have to take the additional overhead of # of futures field. If we specify it is a null-terminated list, then there is no confusion, and it is clear there cannot be nullptr in between. I do not know of any benchmark that would need such a feature of having nullptr in the middle.

srirajpaul:
I also thought of the same after adding the comment. We could keep the default option as enabling arbitrary numbers of futures for the same reasons.
More advanced users can turn it off for performance-related requirements.

agrippa:
@srirajpaul just added the compile-time flag to enable/disable this feature. Please take a look, as it's been a while since I added a new flag to the compile scripts.

srirajpaul:
@agrippa The flag works as expected. Please go ahead and merge.

PR #85, opened 2020-06-18
Fix memory leak caused by waiting_on_extra

agrippa:
@srirajpaul just realized that my recent pull request introduced a memory leak. Could you take a quick look at this fix?

There's one new JSON file added for locales, but that can mostly be ignored.

### DIFF
--- a/locality_graphs/cori_one_worker.json
+++ b/locality_graphs/cori_one_worker.json
@@ -1,0 +1,14 @@
+{
+    "nworkers": 1,
+    "declarations": [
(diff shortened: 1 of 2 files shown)


srirajpaul (approved):
Good to go. Just one suggestion. We can free without the if check is it was initialized to nullptr. But feel free to ignore.

srirajpaul:
Just one suggestion. We can free without the if check. It is safe. But feel free to ignore.

PR #86, opened 2020-06-18
Make cmake the default build option

srirajpaul:
Change the build system from Configure to cmake.
The original configure scripts are retained and can be used with install_orig.sh, clean_orig.sh, README_orig.md.
HClib applications/examples/test will continue to work with the earlier Makefile and also can be used with cmake scripts.


srirajpaul:
@agrippa @ahayashi @Matthew-Whitlock @sbak5 Can you please take a look? This makes cmake the default build option for HClib and HClib_system. The earlier Makefile for applications/test will continue to work. test/cpp also includes a sample CMakeLists.txt that will serve as reference for later conversion of applications.

ahayashi:
Thanks @srirajpaul, the PR looks good to me.

README_orig.md
```diff
@@ -0,0 +1,131 @@
+HClib
```
agrippa: @srirajpaul is there any reason to keep this file around? I'm okay with deleting it, we'll always have this in the git history if we need to change anything.

clean_orig.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
agrippa: @srirajpaul also good with deleting this file.

install_orig.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
```
agrippa: @srirajpaul delete this one too?

agrippa:
@srirajpaul I'm no cmake expert but I'm pretty good with these changes. I'm sure we'll discover places where we want to add CMakeLists.txt files in the future but this is a great start.

Is it possible to rebase all of this in a single commit (once you've addresses/replied to my comments above)?

README_orig.md
srirajpaul: @agrippa Just kept for history. Since it is in git history we can delete this file.

clean_orig.sh
srirajpaul: Deleted.

install_orig.sh
srirajpaul: Deleted.

srirajpaul:
@agrippa  Regarding rebasing. Do you want me to rebase all the commits into a single commit or you want to use the "squash and merge" option in github to do it while merging a pull-request?

agrippa:
@srirajpaul ah, i didn't realize there was a github feature for that. if that's easier then i'm good with that. this looks ready to merge to me.

PR #87, opened 2020-07-17
Add hclib_default_queue_capacity

agrippa:


srirajpaul (approved):
Looks good.

src/hclib-runtime.c
```diff
@@ -477,6 +477,14 @@ static inline void execute_task(hclib_task_t *task) {
     free(task);
 }
 
+void hclib_default_queue_capacity(int* used, int* capacity) {
+    const int wid = hclib_get_current_worker();
+    hclib_locale_t *default_locale = hc_context->graph->locales + 0;
```
srirajpaul, line 482: Why add a 0? Just to understand.

agrippa: i sometimes like to do this to make it clear that we're getting the address of the first element in an array. If it's just hc_context->graph->locales it can be unclear to a reader whether locales is just a pointer to an object that we're getting the address of, or a pointer to an array of objects and we're getting the first element. Personally, I think adding +0 makes it look more like we're indexing in to an array. so you get more semantic information from the code.

PR #89, opened 2020-07-17
Add a std::vector<>* variant of async await, set CRAYPE_LINK_TYPE to install script to make building on Cray machines simpler

agrippa:
@srirajpaul just a couple of small changes related to the herds project. Do you see any problem just setting CRAYPE_LINK_TYPE in the install script?

### DIFF
--- a/inc/hclib-async.h
+++ b/inc/hclib-async.h
@@ -232,6 +232,19 @@ inline void async_nb_await_at(T&& lambda, std::vector<hclib_future_t *> &&future
 }
 
 template <typename T>
+inline void async_nb_await_at(T&& lambda, std::vector<hclib_future_t *> *futures,
+        hclib_locale_t *locale) {
+    async_await_at_helper(lambda, futures ? futures->data() : nullptr,
+            futures ? futures->size() : 0, locale, 1);
+}
+
+template <typename T>
+inline void async_nb_await(T&& lambda, std::vector<hclib_future_t *> *futures) {
+    async_await_at_helper(lambda, futures ? futures->data() : nullptr,
+            futures ? futures->size() : 0, nullptr, 1);
+}
+
+template <typename T>
 inline void async_await(T&& lambda, hclib_future_t *future) {
 	MARK_OVH(current_ws()->id);
     typedef typename std::remove_reference<T>::type U;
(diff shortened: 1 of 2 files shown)


srirajpaul:
@agrippa Adding CRAYPE_LINK_TYPE to install script should not be a problem, since it won't set it directly to the global environment. Only case is if user does 'source install.sh', then it will be added to the global environment, but we do not suggest using source.
Also CRAYPE_LINK_TYPE should not cause any problem in non-cray machines.

inc/hclib-async.h
```diff
@@ -286,6 +299,12 @@ inline void async_await(T&& lambda, std::vector<hclib_future_t *> &&futures) {
     async_await_at_helper(lambda, futures.data(), futures.size(), nullptr, 0);
 }
 
+template <typename T>
+inline void async_await(T&& lambda, std::vector<hclib_future_t *> *futures) {
```
srirajpaul, line 303: Compiler can get confused between this and 'async_await(T&& lambda, hclib_future_t *future)'  (https://github.com/srirajpaul/hclib/blob/f2f48c5aafa51aa3daf4b12cfad8b2e21cc88d0a/inc/hclib-async.h#L235)
 if user passes async_await(..., nullptr).
But if passing nullptr is not supported then it is fine.
In those cases user will have to typecast 'async_await(..., (hclib_future_t *)nullptr)' to select the correct API.

inc/hclib-async.h
```diff
@@ -231,6 +231,19 @@ inline void async_nb_await_at(T&& lambda, std::vector<hclib_future_t *> &&future
     async_await_at_helper(lambda, futures.data(), futures.size(), locale, 1);
 }
 
+template <typename T>
+inline void async_nb_await_at(T&& lambda, std::vector<hclib_future_t *> *futures,
```
srirajpaul, line 235: Same as async_await, operator overloading confusion with nullptr.

inc/hclib-async.h
```diff
@@ -231,6 +231,19 @@ inline void async_nb_await_at(T&& lambda, std::vector<hclib_future_t *> &&future
...
+        hclib_locale_t *locale) {
+    async_await_at_helper(lambda, futures ? futures->data() : nullptr,
+            futures ? futures->size() : 0, locale, 1);
+}
+
+template <typename T>
+inline void async_nb_await(T&& lambda, std::vector<hclib_future_t *> *futures) {
```
srirajpaul, line 242: Same as async_await, operator overloading confusion with nullptr.

inc/hclib-async.h
agrippa: @srirajpaul i think that's a good point, but I'm not sure what to do about it (nor do I think it's a huge problem). I do think that passing a null pointer should be supported, but it doesn't really make a difference -- each API handles it in the same way, by treating it as is no futures have been passed.

In general, I don't expect too many people to do async_await(..., nullptr), because why wouldn't they just be calling async(). Mostly, I expect people to do async_await(..., list_of_futures) where list_of_futures is set to nullptr at some point. That's the important case I think, and in that case the type of list_of_futures would tell the compiler which API is being called. do you agree?

srirajpaul: @agrippa async_await(..., list_of_futures) where list_of_futures is set to nullptr at some point should work since  list_of_futures would tell compiler the type to use. In case some one wants to use async_await(..., nullptr), we can say them that  async() is the preferred option for that. We can go ahead and merge it.

